Issue #1, opened 2023-08-16
Switch from GitLab to GitHub


cameronrutherford:
Before Initial Switch:
- [x] Clean up stale branches / issues (#37) 
- [x] Merge any remaining work in MRs
- [x] Issues moved from GitLab to GitHub (also move any stale PRs manually)
- [x] PNNL GitLab push mirror (#7)
- [x] PNNL GPU pipelines posting status (#7)
- [x] Ascent pipelines posting status (#7)

After Initial Switch:
- [x] Tag new release to signify switch, and update spack package with URL + version (Planned 1.6.0)
  - [x] Ensure tags are updated and all branches match
- [x] Update about section and logo in GitHub (@abhyshr took care of this)
- [x] Add Issue / Merge request template (#36)
- [x] Badges updated for pipelines / ExaGO logo update (#15)
  - [x] This includes the new GitHub actions, and pre-commit (maybe codecov one day) [![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit)](https://github.com/pre-commit/pre-commit)
- [x] Crusher pipelines fixed and posting status. (#9)
- [x] GitHub Actions pipelines
  - [x] pre-commit instead of CMake based linting (#5)
  - [x] spack CI builds in GitHub actions for CPU only builds (#11 #15)

Follow up work:
- [x] Quarto Documentation + GitHub wiki integration (#32) 
- [x] Crusher CI status posting (#9) 
- [x] Add spack module pipelines to Ascent and Crusher (#10 #34) 
- [x] Submodules removed in favor of external packages (#38)

cameronrutherford:
pre-commit documentation is a must

cameronrutherford:
In addition to MR / Issue template, we should include requirements to manually run and have crusher CI pass before merging. Can't enforce through GitHub, so this is a manual / optional req

cameronrutherford:
Captured reamining tasks in sub issues

PR #5, opened 2023-09-14
Add GitHub actions for push mirror

cameronrutherford:
Moves CMake format, Clang format and Python linting into pre-commit, as well as moving the file naming conventions check into GitHub actions.

This also removes the checks from the GitLab pipelines and removes the old perl code that is handled by pre-commit now.

Related to #1


cameronrutherford:
cc @ryandanehy as I wasn't able to add you as a reviewer.

cameronrutherford:
There are a few more changes that I want to make, but this is an atomic amount of change that I wanted to start with

.github/workflows/develop_main_mirror.yaml
```diff
@@ -0,0 +1,30 @@
...
+     runs-on: ubuntu-22.04
+     steps:
+       # Action checks-out your repository under $GITHUB_WORKSPACE, so your workflow can access it.
+       - uses: actions/checkout@v1
+       #  Action for mirroring your commits to a different remote repository
+       - uses: yesolutions/mirror-action@master
+       # Give actions access to some secrets
+         with:
+            GIT_PUSH_ARGS: '--tags --force'
+            REMOTE: 'https://gitlab.pnnl.gov/exasgd/frameworks/exago-github-mirror'
```
jaelynlitz, line 23: Does this repo exist? Should I be able to see it?

buildsystem/tools/lib/ExaGO.pm
```diff
@@ -12,10 +12,7 @@ use warnings;
 use v5.16;
 
 use ExaGO::Env;
-use ExaGO::ClangFormat;
-use ExaGO::CMakeFormat;
 use ExaGO::FileNamingConventions;
```
jaelynlitz, line 15: This one too?

tests/unit/opflow/gradient/gen_network.py
```diff
@@ -2,115 +2,137 @@
...
+            f"{bus_idx} {bus_idx + 1} 0.000000 0.000010 0.00000 0.00 0.00 0.00 0.00000 0.000 1 0.00 0.00 0.00 0.00 0.00 0.00\n")
+        output.append(
+            f"{bus_idx + 1} {bus_idx + 2} 0.000000 0.000010 0.00000 0.00 0.00 0.00 1.00000 0.000 1 0.00 0.00 -10.00 -0.00 10.00 0.00\n")
+        output.append(
+            f"{bus_idx + 1} {bus_idx + 3} 0.000000 0.000010 0.00000 0.00 0.00 0.00 0.00000 0.000 1 0.00 0.00 10.00 10.00 -10.00 -10.00\n")
+        output.append(
+            f"{bus_idx + 3} {bus_idx + 4} 0.000000 0.000010 0.00000 0.00 0.00 0.00 0.00000 0.000 1 0.00 0.00 0.00 0.00 0.00 0.00\n")
+    output.append("];\n")
+    output.append("\n")
+    return output
```
jaelynlitz, line 83: These are all just formatting changes right? And precommit makes these changes itself upon push?

buildsystem/tools/lib/ExaGO.pm
jaelynlitz: Ah nvm I see that the yaml still calls this perl script?

jaelynlitz (approved):
LGTM

.github/workflows/develop_main_mirror.yaml
cameronrutherford: Yeah it's there. Not sure who all will need access but it should be given to everyone a part of the ExaSGD group on GitLab I would hope

buildsystem/tools/lib/ExaGO.pm
cameronrutherford: Yeah so the clang-format, cmake-format and autopep8 pipelines were super trivial, but the naming conventions one I had a lot more difficulty in fixing. I tried a whole bunch of things with regex but I am not wise enough in the ways of Python globbing and the configuration to make it worth the effort. For now I decided to keep naming conventions as-is

tests/unit/opflow/gradient/gen_network.py
cameronrutherford: Yeah all formatting changes. These files weren't linted at all before so that's why these are being touched for the first time.

pre-commit runs a check when you run `git commit`, and then applies changes to files as needed. You then re-add them, run `git commit` again, and pre-commit checks will be a noop and your commit will go through just fine.

PR #13, opened 2023-09-19
Logger fix

jaelynlitz:
enable logging by default on incline and crusher, fix bug in logger test to now only run if logging is enabled


cameronrutherford (dismissed):
Closes https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/issues/460

cameronrutherford (approved):

PR #14, opened 2023-09-20
Make Github ignore Matlab (*.m) files when detecting programming languanges

abhyshr:
Github shows the main programming language for ExaGO as Matlab. This is because there are a lot of data files with .m extension. This PR adds a .gitattributes file that tells Github's language checker (linguist) to ignore all .m files. Solution from https://dev.to/katkelly/changing-your-repo-s-language-in-github-5gjo


cameronrutherford:
pre-commit is failing. There should be a way to configure this pipeline to apply fixes where possible. Let me look into that briefly but we should be able to merge this very quickly

PR #15, opened 2023-09-22
Spack GitHub action

### LINKED ISSUE #11: CPU only build of ExaGO in GitHub actions
Similar to how HiOp does this, we should have some ExaGO build(s) happen in CI that are CPU only.

This should be a minimal build (no testing, no logging, no GPU), and so should be quick to build, and will catch some more of the xSDK related build issues such as https://github.com/xsdk-project/xsdk-issues/issues/220#issuecomment-1719971677

cameronrutherford:
This modifies exago's spack package, so https://github.com/spack/spack/pull/40188 should be merged before this, and submodule updated accordingly.

Closes #11


.github/workflows/spack_cpu_build.yaml
```diff
@@ -0,0 +1,52 @@
...
+    container: spack/ubuntu-focal:latest
+    strategy:
+      matrix:
+        spack_spec:
+          - exago@develop+mpi~ipopt~hiop~logging~python~raja ^openmpi
+          - exago@develop~mpi~ipopt+hiop~logging~python~raja
+          - exago@develop~mpi~ipopt~hiop~logging~python~raja
+          - exago@develop~mpi~ipopt~hiop~logging+python~raja
+          - exago@develop~mpi~ipopt+hiop~logging~python+raja
+          - exago@develop~mpi~ipopt~hiop~logging~python+raja
```
cameronrutherford, line 19: @abhyshr we are going to build out this matrix through things like Spack CI and connecting GitLab runners to AWS, so this is just a first pass, but what do you think of the usefulness of this set of builds? Are there any that you would cut?

cc @Jayapreethi and @paulrigor

.github/workflows/spack_cpu_build.yaml
```diff
@@ -0,0 +1,52 @@
...
+          # Need to create an environment to install hiop in the action's branch
+          spack env create -d ./spack-env
+          spack env activate ./spack-env
+          spack add $SPACK_SPEC target=x86_64
+          spack develop --path $(pwd) --no-clone exago@git.${{ github.event.pull_request.head.ref}}=develop
+          # Need to add build cache before concretizing to re-use dependencies
+          # Add E4S mirror - likely relying on spack cache but nice backup
+          # https://oaciss.uoregon.edu/e4s/inventory.html
+          spack mirror add E4S https://cache.e4s.io
+          spack buildcache keys --install --trust
```
cameronrutherford, line 49: We trust two unknown sets of keys here, and so this also further motivates private AWS pipelines in order to control our own ecosystems. The caches are still useful in this context, but it would also make our lives easier to just use our own cache that serves all of our needs securely instead of this.

cameronrutherford (commented):
While we still have other work to do with submodules, these builds should catch some errors we are seeing in xSDK pipelines and so are useful in their current state. This will be improved in the future.

cameronrutherford:
Currently each of the spack builds take at least 30min of GitHub runner time, so we should both minimize the set of configurations we test, and also try to only run CI infrequently (maybe only on merges into develop from now on?)

.github/workflows/spack_cpu_build.yaml
abhyshr: For all our scopflow and sopflow runs, IPOPT is a mandatory requirement. So, I think you can condense this to 
+ipopt+python
+ipopt+hiop+python
+ipopt+hiop+raja+python

is there a reason you want also want to do tests without mpi? All our applications use MPI so why the need for no MPI tests?

We also need testing with different optimization options (debug, release)

cameronrutherford: These tests aren't actually running anything at the moment. In the future, we might do a small `spack test exago`, but for now this is just to catch build failures in configurations currently not built in CI.

We can build everything with MPI, but that would drastically increase the build size, and so I don't want more than one MPI build here for the moment without good reason.

We can't do `+ipopt` in CI here since we would need the CoinHSL tarball in GitHub actions/our repo, which is against the license. So we can't have any of the builds here be `+ipopt`.

cameronrutherford: Building with a few different release types is a good suggestion. I'll try and add that on Monday

cameronrutherford:
https://github.com/pnnl/ExaGO/actions/runs/6279332431/job/17054811101?pr=15 - really interesting bug here. This is probably one of the first times that someone has built exago in this minimal config, so we have an error that we don't `import os` in our package.

Normally something like `hiop` or another package would conveniently import this for us, but we finally hit a build configuration where this wasn't the case :)

See spack PR for the fix that is used in this PR (ideally Spack PR is merged before this) https://github.com/spack/spack/pull/40188

cameronrutherford:
https://github.com/xsdk-project/xsdk-issues/issues/220 - finally reproduced this issue in the pipelines here. Just a matter of actually fixing the issue now.

cameronrutherford:
@ryandanehy and @abhyshr I think this is ready to go when CI passes, so I would love at least one more review of this finished product before merging.

This also resolves issues with xSDK build failures, so good to get this merged soon.

cameronrutherford:
See #17 for issue identified when building `exago~tests`

cameronrutherford:
See #18 and https://github.com/LLNL/hiop/issues/662 for issues with `exago~mpi+hiop`

abhyshr (approved):
Looks good to me.

Issue #18, opened 2023-09-25
`exago+hiop~ipopt+logging~mpi~python+raja+tests`


cameronrutherford:
https://github.com/LLNL/hiop/issues/662 - hard to track down the root cause here, but I suspect this is a hiop issue in the end.

Core config that catches this bug seems to be `+hiop~mpi`

cameronrutherford:
#15 is where this was identified

cameronrutherford:
`~hiop~mpi` also fails as a build configuration, and is fixed in https://github.com/spack/spack/pull/40188

Documenting this here for now as another `~mpi` failing build.

cameronrutherford:
Moving to 1.6.2 since build `~mpi` are low priority atm

PR #21, opened 2023-09-29
Update summit modules

nkoukpaizan:
This PR updates the Summit modules.

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -13,13 +13,13 @@ jobs:
       matrix:
         spack_spec:
           # Minimal Build
-          - exago@develop~mpi~ipopt~hiop~logging~python~raja~tests
-          - exago@develop+mpi~ipopt+hiop~logging~python~raja~tests ^openmpi
+          - exago@develop~mpi~ipopt~hiop~python~raja
+          - exago@develop+mpi~ipopt+hiop~python~raja ^openmpi
           # See #18 - +hiop~mpi causes issues
-        # - exago@develop~mpi~ipopt+hiop+logging~python+raja+tests
+        # - exago@develop~mpi~ipopt+hiop~python+raja
           # See #16 - +python~mpi causes issues
-        # - exago@develop~mpi~ipopt+hiop~logging+python~raja+tests
-          - exago@develop+mpi~ipopt+hiop~logging+python~raja+tests ^openmpi
+        # - exago@develop~mpi~ipopt+hiop+python~raja
+          - exago@develop+mpi~ipopt+hiop+python~raja ^openmpi
 
     name: Build ExaGO with Spack
     steps:
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -13,7 +13,8 @@ jobs:
       matrix:
         spack_spec:
           # Minimal Build
-          - exago@develop~mpi~ipopt~hiop~python~raja
+           # See #18 
+        # - exago@develop~mpi~ipopt~hiop~python~raja
           - exago@develop+mpi~ipopt+hiop~python~raja ^openmpi
           # See #18 - +hiop~mpi causes issues
         # - exago@develop~mpi~ipopt+hiop~python+raja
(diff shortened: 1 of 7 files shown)


cameronrutherford (approved):

cameronrutherford:
Even with changes removing the `tests` and `logging` variant, we are still blocked because the non-mpi build needs to import `os`.

@nkoukpaizan if you want this merged ASAP, we can disable the `~mpi` build for now, and re-add once Spack PR is merged.

.github/workflows/spack_cpu_build.yaml
```diff
@@ -13,13 +13,13 @@ jobs:
       matrix:
         spack_spec:
           # Minimal Build
-          - exago@develop~mpi~ipopt~hiop~logging~python~raja~tests
-          - exago@develop+mpi~ipopt+hiop~logging~python~raja~tests ^openmpi
+          - exago@develop~mpi~ipopt~hiop~python~raja
```
cameronrutherford, line 16:
```suggestion
           # See #18 
        # - exago@develop~mpi~ipopt~hiop~python~raja
```

cameronrutherford (approved):
#18

cameronrutherford:
@ryandanehy or @nkoukpaizan please give this a quick review as GitHub is requiring an additional reviewer (since I made the latest changes).

cameronrutherford:
Need to re-visit but it seems like Python CMake configuration is erroring when Python is not found. Simple problem but need to make sure we make a robust solution.

cameronrutherford:
Spack PR was merged, so I think if we update the submodule to include these changes the pipelines will pass

nkoukpaizan:
Python build is failing. I think we need to update the pybind11 submodule. This closed issue in their repository seems relevant: https://github.com/pybind/pybind11/issues/4464

cameronrutherford:
@nkoukpaizan I am forever thankful that you were able to reproduce #23 :)

cc @jaelynlitz as she was going down a rabbit hole of updating the compiler to catch this bug, but it seems like we don't quite need to do that. @jaelynlitz please only modify compilers on CI platforms if you need to do so in order to upgrade C++ standard.

@nkoukpaizan I agree with your assessment that we need to change our pybind submodule to fix this. We need to also move pybind dep into spack eventually, but this is an easy thing to fix for the moment.

cameronrutherford:
> @nkoukpaizan I am forever thankful that you were able to reproduce #23 :)
> 
> cc @jaelynlitz as she was going down a rabbit hole of updating the compiler to catch this bug, but it seems like we don't quite need to do that. @jaelynlitz please only modify compilers on CI platforms if you need to do so in order to upgrade C++ standard.
> 
> @nkoukpaizan I agree with your assessment that we need to change our pybind submodule to fix this. We need to also move pybind dep into spack eventually, but this is an easy thing to fix for the moment.

Since this fixes CI, I think this also closes #23

jaelynlitz:
> @nkoukpaizan I am forever thankful that you were able to reproduce #23 :)
> 
> cc @jaelynlitz as she was going down a rabbit hole of updating the compiler to catch this bug, but it seems like we don't quite need to do that. @jaelynlitz please only modify compilers on CI platforms if you need to do so in order to upgrade C++ standard.
> 
> @nkoukpaizan I agree with your assessment that we need to change our pybind submodule to fix this. We need to also move pybind dep into spack eventually, but this is an easy thing to fix for the moment.

What exactly was the fix? updating pybind? I also see a magma upgrade from 2.6.2 to 2.7.1?

cameronrutherford:
> > @nkoukpaizan I am forever thankful that you were able to reproduce #23 :)
> 
> > 
> 
> > cc @jaelynlitz as she was going down a rabbit hole of updating the compiler to catch this bug, but it seems like we don't quite need to do that. @jaelynlitz please only modify compilers on CI platforms if you need to do so in order to upgrade C++ standard.
> 
> > 
> 
> > @nkoukpaizan I agree with your assessment that we need to change our pybind submodule to fix this. We need to also move pybind dep into spack eventually, but this is an easy thing to fix for the moment.
> 
> 
> 
> What exactly was the fix? updating pybind? I also see a magma upgrade from 2.6.2 to 2.7.1?

Fix to Python error was pybind. Magma upgrade might cause testing errors, but these are summit modules not run in CI, so it's harder to track.

cameronrutherford (approved):

PR #24, opened 2023-10-02
tests on multiple mpi ranks

bjpalmer:
The func-test-mpi branch has been rebased against develop and appears to be ready to merge. This PR closes issue #27.


cameronrutherford:
All the commits were from @jaelynlitz - is this her previous work?

I also would appreciate if you could bring over the relevant issue from GitLab, close that issue, and then ensure that this PR closes that issue upon merge.

bjpalmer:
There doesn't appear to be an issue in Gitlab, unless I'm being obtuse. @jaelynlitz do you see it? I may have picked this up by scanning through the branches.

cameronrutherford:
> There doesn't appear to be an issue in Gitlab, unless I'm being obtuse. @jaelynlitz do you see it? I may have picked this up by scanning through the branches.

https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/issues/279

https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/merge_requests/269

https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/merge_requests/472/diffs

https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/merge_requests/293

https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/issues/284

Does this resolve the SCOPFLOW tests printing duplicate output perhaps? I am confused as there are two existing PRs which have been merged targeting pflow, and this would be the third.

It would be great if you could consolidate the open issue into individual components, and clarify what in particular this PR is adressing.

tests/functionality/toml_utils.hpp
```diff
@@ -0,0 +1,230 @@
+#pragma once
```
cameronrutherford, line 1: Erroneously named (toml_utils.hpp), as we already have something called `toml_utils.h`

bjpalmer: toml_utils.hpp doesn't appear to be used anywhere and deleting it doesn't appear to have any effect on the test suite. I also did a diff on toml_utils.hpp and toml_utils.h and toml_utils.h appears to have more recent updates so I think it is safe to get rid of toml_utils.hpp. I will do this on the func-test-mpi branch.

bjpalmer: The first issue has been close and the second and third merge PRs have been merged. The fourth PR is duplicated in Github by this PR so the only outstanding issue is 284 in Gitlab. I will copy this issue to Github.

cameronrutherford: Okay so is there a specific change that this PR is addressing, or does more work need to be done in order to address SCOPFLOW/SOPFLOW parallel tests?

cameronrutherford: Thanks for cleaning this up! Next steps then are to implement the fixes in `selfcheck.cpp` for both SOPFLOW and SCOPFLOW, and then we can merge.

cameronrutherford:
Tentatively marking for 1.6.1 release unless this is about to be merged

bjpalmer:
If you are otherwise ready to go, I'd skip this one. There still seem to be some issues.

bjpalmer:
Attached is a log file from Valgrind when running the FUNCTIONALITY_TEST_SCOPFLOW_HIOP_SERIAL_TESTSUITE test in the test suite. This corresponds to the hiop_serial.toml file under tests/functionality/scopflow.

[log.2203724.txt](https://github.com/pnnl/ExaGO/files/12895674/log.2203724.txt)

bjpalmer:
I looked through the code to see if I can track down some of the error showing up in Valgrind. I can make the errors coming from PetscMemcpy go away by using a char* buffer and strncpy even though I'm not completely sure why the original error is showing up.

Most of the remaining errors are complaints that the following three variables in OPFLOW are not initialized:

```
ignore_lineflow_constraints
include_powerimbalance_variables
solver.default_value
```

@abhyshr, what are appropriate default values for these variables and where should they be set?

bjpalmer:
Test 34 is now running correctly on Newell after getting rid of the error reported from PetscMemcpy, but I think it would still be a good idea to make sure the variables above are initialized to something.

bjpalmer:
I looked in the code in the opflow.h file and it appears that both ignore_lineflow_constraints and include_powerimbalance_variables should be initialized to false. Maybe valgrind doesn't like ExaGOBoolOption function?

bjpalmer:
I was able to track down some of the uninitialized variables. The variables are initialized in the SCOPFLOW and OPFLOW Initialization routines but then are overwritten in the selfcheck.cpp file with variables that may not have been initialized. The failure of test 31 on Newell and Deception disappeared but new failures for tests 34 and 35 showed up. @abhyshr, can you take a look at these failures? The test 34 failure is a largish change in the objective function and the number of iterations going from 1 to 2. Is it possible that this may actually be an improved answer?

cameronrutherford:
> I was able to track down some of the uninitialized variables. The variables are initialized in the SCOPFLOW and OPFLOW Initialization routines but then are overwritten in the selfcheck.cpp file with variables that may not have been initialized. The failure of test 31 on Newell and Deception disappeared but new failures for tests 34 and 35 showed up. @abhyshr, can you take a look at these failures? The test 34 failure is a largish change in the objective function and the number of iterations going from 1 to 2. Is it possible that this may actually be an improved answer?

Are the test failures in CI? Would be easiest to link logs there...

bjpalmer:
The test 34 failure is showing up on both Newell and Deception in CI.

abhyshr:
I am not sure if we have changed anything in SOPFLOW that will affect the solution. I don't recall of anything at this point. Not sure if it has to do with any of the dependency libraries. Did something change in the input files. I am not sure. Might be best to update the reference solution for the two tests for now and see if we are get any more failures in the future.

bjpalmer:
I tried rebasing against develop and got a conflict with .github/workflows/spack_cpu_build.yaml but there is no .github directory listed in the top level exago directory. How am I getting a conflict here?

cameronrutherford:
> I tried rebasing against develop and got a conflict with .github/workflows/spack_cpu_build.yaml but there is no .github directory listed in the top level exago directory. How am I getting a conflict here?

I assume that you are not able to view the `.github` directory in your terminal when doing this, as that folder is in good health in `exago@develop`.

It appears as though your git history is mangled, so I suggest the generic approach assuming you are in a checked out local version of `func-test-mpi`, and no existing `develop` branch has been checked out locally:

```
$ git fetch --all
$ git checkout -b develop --track origin/develop
$ git reset --hard origin/develop
$ git checkout -
$ git rebase -i develop
$ git push -f origin <branch_name>
```

bjpalmer:
Is that second line correct? The develop branch already exists.

cameronrutherford:
> Is that second line correct? The develop branch already exists.

> assuming you are in a checked out local version of func-test-mpi, and no existing develop branch has been checked out locally

bjpalmer:
Sorry, I meant the second git instruction

```
$ git checkout -b develop --track origin/develop
```

I'm getting a complaint about the -b.

bjpalmer:
I just rebased against develop again this morning. If everything works, I'd merge before anything else changes.

src/opflow/interface/opflowregi.cpp
```diff
@@ -76,10 +76,11 @@ extern PetscErrorCode OPFLOWModelCreate_PBPOLHIOP(OPFLOW);
 
 #if defined(EXAGO_ENABLE_RAJA)
 extern PetscErrorCode OPFLOWModelCreate_PBPOLRAJAHIOP(OPFLOW);
+#endif
+
 #if defined(EXAGO_ENABLE_HIOP_SPARSE)
 extern PetscErrorCode OPFLOWModelCreate_PBPOLRAJAHIOPSPARSE(OPFLOW);
 #endif
-#endif
```
cameronrutherford: This is incorrect. Please revert this change

src/opflow/interface/opflowregi.cpp
```diff
@@ -117,11 +118,12 @@ PetscErrorCode OPFLOWModelRegisterAll(OPFLOW opflow) {
...
                              OPFLOWModelCreate_PBPOLRAJAHIOP);
   CHKERRQ(ierr);
+#endif
+
 #if defined(EXAGO_ENABLE_HIOP_SPARSE)
   ierr = OPFLOWModelRegister(opflow, OPFLOWMODEL_PBPOLRAJAHIOPSPARSE,
                              OPFLOWModelCreate_PBPOLRAJAHIOPSPARSE);
   CHKERRQ(ierr);
-#endif
 #endif
```
cameronrutherford: This is also incorrect. Please revert this

src/opflow/interface/opflowregi.cpp
```diff
@@ -134,13 +136,13 @@ extern PetscErrorCode OPFLOWSolverCreate_IPOPT(OPFLOW);
...
 extern PetscErrorCode OPFLOWSolverCreate_HIOP(OPFLOW);
+#endif
 #if defined(EXAGO_ENABLE_HIOP_SPARSE)
 extern PetscErrorCode OPFLOWSolverCreate_HIOPSPARSE(OPFLOW);
+#endif
 #if defined(EXAGO_ENABLE_RAJA)
 extern PetscErrorCode OPFLOWSolverCreate_HIOPSPARSEGPU(OPFLOW);
 #endif
-#endif
-#endif
```
cameronrutherford: Please revert these changes

src/utils/utils.cpp
```diff
@@ -584,12 +584,19 @@ PetscErrorCode ExaGOInitialize(MPI_Comm comm, int *argc, char ***argv,
...
+  if (err)
+    throw ExaGOError("Error getting MPI rank number");
   bool flg;
   ExaGOLogIsUsingLogFile(&flg);
   if (flg) {
     std::string filename;
     ExaGOLogGetLoggerName(filename);
-    ExaGOLog(EXAGO_LOG_INFO, "See logfile {} for output.", filename);
+    if (my_rank == logging_rank)
+      ExaGOLog(EXAGO_LOG_INFO, "See logfile {} for output.", filename);
```
cameronrutherford: I don't know if it's really necessary to store logging_rank as a variable here - why not just hardcode this to be `if (my_rank == 0)`?

tests/functionality/opflow/selfcheck.cpp
```diff
@@ -11,11 +11,12 @@ struct OpflowFunctionalityTestParameters {
   std::string solver = "";
   std::string model = "";
   std::string network = "";
-  bool is_loadloss_active;
-  bool is_powerimb_active;
-  std::string gen_bus_voltage_string;
+  bool is_loadloss_active = false;
+  bool is_powerimb_active = false;
```
cameronrutherford, line 12: Do these also have OPFLOW default values?

tests/functionality/opflow/selfcheck.cpp
```diff
@@ -24,8 +25,9 @@ struct OpflowFunctionalityTestParameters {
   double power_imbalance_penalty = 1000.0;
   std::string description = "";
   int hiop_verbosity_level = 0;
-  std::string hiop_compute_mode;
-  std::string hiop_mem_space;
+  std::string hiop_compute_mode =
+      OPFLOWOptions::hiop_compute_mode.default_value;
+  std::string hiop_mem_space = OPFLOWOptions::hiop_mem_space.default_value;
```
cameronrutherford, line 27: I don't immediately see anything wrong with these changes, but is there a motivation for default initializing these? How did this work beforehand?

tests/functionality/pflow/selfcheck.cpp
```diff
@@ -34,52 +34,76 @@ struct PflowFunctionalityTests
...
+      throw ExaGOError("Error getting MPI rank number");
+
     auto err = MPI_Comm_size(comm, &nprocs);
-    if (err)
-      throw ExaGOError("Error getting MPI num ranks");
+    if (err) {
+      if (my_rank == logging_rank)
+        throw ExaGOError("Error getting MPI num ranks");
+      exit(0);
+    }
```
cameronrutherford: Again I don't think `logging_rank` is relevant here as it's just a 0

tests/functionality/pflow/selfcheck.cpp
```diff
@@ -34,52 +34,76 @@ struct PflowFunctionalityTests
...
-           << presets;
-      throw ExaGOError(errs.str().c_str());
+      if (my_rank == logging_rank) {
+        std::stringstream errs;
+        errs << "Number of processes should be declared globally in the preset "
+                "area of the test suite TOML file, not inside each testcase.\n"
+             << "Testcase: " << testcase << "\nWith presets:\n"
+             << presets;
+        throw ExaGOError(errs.str().c_str());
+      }
```
cameronrutherford: Same here, logging_rank is just 0

tests/functionality/pflow/selfcheck.cpp
```diff
@@ -34,52 +34,76 @@ struct PflowFunctionalityTests
...
-             << " configuration:\n";
-        errs << testcase << "\nwith these presets:\n" << presets;
-        throw ExaGOError(errs.str().c_str());
+        if (my_rank == logging_rank) {
+          std::stringstream errs;
+          errs << "PFLOW Test suite expected option '" << opt
+               << "' to be available, but it was not found in this testsuite"
+               << " configuration:\n";
+          errs << testcase << "\nwith these presets:\n" << presets;
+          throw ExaGOError(errs.str().c_str());
```
cameronrutherford: Throwing an error here should be sufficient. Since these are void functions as well, I don't think you need `exit(0)` in any of these functions.

tests/functionality/pflow/selfcheck.cpp
```diff
@@ -108,23 +132,26 @@ struct PflowFunctionalityTests
...
+    if (my_rank == logging_rank)
       std::cout << "Test Description: " << params.description << std::endl;
-    }
     ierr = PFLOWCreate(params.comm, &pflow);
     ExaGOCheckError(ierr);
 
     // Prepend installation directory to network path
     resolve_datafiles_path(params.network);
+    strncpy(pbuf, params.network.c_str(), params.network.length());
+    pbuf[params.network.length()] = '\0';
```
cameronrutherford: This should either be a part of our API, or removed

tests/functionality/pflow/selfcheck.cpp
```diff
@@ -192,10 +219,16 @@ int main(int argc, char **argv) {
   ExaGOCheckError(ierr);
   ExaGOLog(EXAGO_LOG_INFO, "{}", "Creating PFlow Functionality Test");
 
+  int my_rank;
+  auto err = MPI_Comm_rank(comm, &my_rank);
+  if (err)
+    throw ExaGOError("Error getting MPI rank number");
+
+  if (my_rank == 0)
```
cameronrutherford, line 193: You use 0 here...

tests/functionality/scopflow/CMakeLists.txt
```diff
@@ -136,7 +136,7 @@ if(EXAGO_INSTALL_TESTS)
     COMMAND
     ${MPICMD}
     "-n"
-    3
+    1
```
cameronrutherford, line 135: I am suspicious as this seems like an unnecessary change

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -8,35 +8,39 @@ struct ScopflowFunctionalityTestParameters {
...
-  std::string opflow_genbusvoltage_string;
+  std::string opflow_genbusvoltage_string =
+      OPFLOWOptions::genbusvoltage.default_value;
   int opflow_genbusvoltage;
-  std::string subproblem_solver;
-  std::string subproblem_model;
-  bool enable_powerimbalance_variables;
-  bool ignore_lineflow_constraints;
-  int verbosity_level;
+  std::string subproblem_solver = OPFLOWOptions::solver.default_value;
```
cameronrutherford, line 35: Again I question if these default initializations are actually necessary. Is there anything that motivated these changes? `develop` seems to have been chugging along just fine without it for a while

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -104,23 +108,81 @@ struct ScopflowFunctionalityTestParameters {
...
 struct ScopflowFunctionalityTests
     : public FunctionalityTestContext<ScopflowFunctionalityTestParameters> {
-  ScopflowFunctionalityTests(std::string testsuite_filename,
-                             int logging_verbosity = EXAGO_LOG_INFO)
-      : FunctionalityTestContext(testsuite_filename, logging_verbosity) {}
 
   using Params = ScopflowFunctionalityTestParameters;
+  MPI_Comm comm;
+  int nprocs;
+  int logging_rank = 0;
```
cameronrutherford: Just use 0...

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -104,23 +108,81 @@ struct ScopflowFunctionalityTestParameters {
...
+
   void
   ensure_options_are_consistent(toml::value testcase,
                                 toml::value presets = toml::value{}) override {
+    int my_rank;
+    auto err = MPI_Comm_rank(comm, &my_rank);
+    if (err)
+      throw ExaGOError("Error getting MPI rank number");
+
+#if 0
```
cameronrutherford: This seems like a remnant from debugging?

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -104,23 +108,81 @@ struct ScopflowFunctionalityTestParameters {
...
-        throw ExaGOError(errs.str().c_str());
+        if (my_rank == logging_rank) {
+          std::stringstream errs;
+          errs << "SCOPFLOW Test suite expected option '" << opt
+               << "' to be available, but it was not found in this testsuite"
+               << " configuration:\n";
+          errs << testcase << "\nwith these presets:\n" << presets;
+          throw ExaGOError(errs.str().c_str());
+        }
+        exit(0);
```
cameronrutherford: Again void functions don't need exit? I think all ranks perhaps then should throw the error, but only one rank should print anything?

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -191,13 +253,14 @@ struct ScopflowFunctionalityTests
...
+    char pbuf[PETSC_MAX_PATH_LEN];
     int rank;
+    auto err = MPI_Comm_rank(comm, &rank);
+    if (err)
+      throw ExaGOError("Error getting MPI rank number");
 
-    MPI_Comm_rank(PETSC_COMM_WORLD, &rank);
-
-    if (rank == 0) {
+    if (rank == logging_rank)
```
cameronrutherford: Just use 0... Maybe just define value in header file if you really want to configure this.

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -191,13 +253,14 @@ struct ScopflowFunctionalityTests
   void run_test_case(Params &params) override {
     PetscErrorCode ierr;
     SCOPFLOW scopflow;
+    char pbuf[PETSC_MAX_PATH_LEN];
```
cameronrutherford: Again I don't think these should live outside of API

tests/functionality/sopflow/selfcheck.cpp
```diff
@@ -112,23 +112,80 @@ struct SopflowFunctionalityTestParameters {
...
 struct SopflowFunctionalityTests
     : public FunctionalityTestContext<SopflowFunctionalityTestParameters> {
-  SopflowFunctionalityTests(std::string testsuite_filename,
-                            int logging_verbosity = EXAGO_LOG_INFO)
-      : FunctionalityTestContext(testsuite_filename, logging_verbosity) {}
 
   using Params = SopflowFunctionalityTestParameters;
+  MPI_Comm comm;
+  int nprocs;
+  int logging_rank = 0;
```
cameronrutherford: ...

tests/functionality/sopflow/selfcheck.cpp
```diff
@@ -112,23 +112,80 @@ struct SopflowFunctionalityTestParameters {
...
+
   void
   ensure_options_are_consistent(toml::value testcase,
                                 toml::value presets = toml::value{}) override {
+    int my_rank;
+    auto err = MPI_Comm_rank(comm, &my_rank);
+    if (err)
+      throw ExaGOError("Error getting MPI rank number");
+
+#if 0
```
cameronrutherford: Another `#if 0`

tests/functionality/sopflow/selfcheck.cpp
```diff
@@ -112,23 +112,80 @@ struct SopflowFunctionalityTestParameters {
...
-             << " configuration:\n";
-        errs << testcase << "\nwith these presets:\n" << presets;
-        throw ExaGOError(errs.str().c_str());
+        if (my_rank == logging_rank) {
+          std::stringstream errs;
+          errs << "SOPFLOW Test suite expected option '" << opt
+               << "' to be available, but it was not found in this testsuite"
+               << " configuration:\n";
+          errs << testcase << "\nwith these presets:\n" << presets;
+          throw ExaGOError(errs.str().c_str());
```
cameronrutherford: No exit(0) here in comparison...

tpl/pybind11
cameronrutherford, line 1: I don't think this should be updated

tpl/spack
cameronrutherford, line 1: Unless there is a need within this PR I don't think this should be updated

cameronrutherford (changes requested):
A lot of these comments are repeats from across applications/tests

tests/functionality/opflow/selfcheck.cpp
bjpalmer: Not really. There are some related variables (e.g. "include_loadloss_variables") that have defaults that could be use to infer that default value, but nothing direct.

src/opflow/interface/opflowregi.cpp
bjpalmer: Don't know how those slipped in. I though I chacked that everything built before I checked the code in. It's fixed now.

src/opflow/interface/opflowregi.cpp
bjpalmer: Done.

src/opflow/interface/opflowregi.cpp
bjpalmer: Done.

tests/functionality/opflow/selfcheck.cpp
cameronrutherford: Don't know the LoE but it would be nice to have these determine their default values in the same way as other options

tests/functionality/sopflow/selfcheck.cpp
bjpalmer: Never mind. I saw this before seeing the comments up above.

I put an exit in, but on looking at the code more carefully, I'm puzzled by the logic here. If you throw an exception, then you would expect that to get caught by the calling program so I'm not sure what exit(0) is supposed to do. I guess if it doesn't get caught then it should exit?

tests/functionality/opflow/selfcheck.cpp
bjpalmer: These values overwrite the default values in the OPFLOW module. If you don't initialize these values to something and they aren't included in the .toml file (which would cause them to be initialized to a specific value), then you are using randomly initialized data to start the calculation off. This will cause problems.

tests/functionality/scopflow/CMakeLists.txt
bjpalmer: The name of this test is FUNCTIONALITY_TEST_SCOPFLOW_HIOP_SERIAL_TESTSUITE so it seems like it should be run on 1 process.

tests/functionality/scopflow/selfcheck.cpp
bjpalmer: This is probably set in all the .toml files. The ones that aren't set likely end up being zero most of the time. The fact that we have been chugging along fine may just mean that we've been lucky. Plus one test started passing when I made these changes (although 2 different ones started failing).

tests/functionality/sopflow/selfcheck.cpp
bjpalmer: Done.

tests/functionality/scopflow/selfcheck.cpp
bjpalmer: Gone.

tests/functionality/sopflow/selfcheck.cpp
cameronrutherford: I think if rank0 (or any rank) throws an exception, then all ranks will exit.

Exit(0) is likely harmless, but we have graceful cleanup in ExaGO utilities, so I think this is unnecessary

tests/functionality/pflow/selfcheck.cpp
bjpalmer: I don't think I was the one that originally put in logging_rank (I could be wrong, my memory is going). @jaelynlitz did you add this? It looks like some kind of future-proofing to me.

tests/functionality/pflow/selfcheck.cpp
bjpalmer: Okay, let me fool around with this one. I did this to get rid of some memory errors reported by valgrind. I'm not sure whether the c_str() method returns a null-terminated string or just a string that is length() characters long.

tests/functionality/pflow/selfcheck.cpp
cameronrutherford: IMO just a 0 everywhere is fine, but we could make a global variable if we really want

tests/functionality/pflow/selfcheck.cpp
cameronrutherford: Maybe related to #31, but we desperately need to use std strings, std filesystem and use C++17.

cc @jaelynlitz

tests/functionality/opflow/selfcheck.cpp
bjpalmer: I could add these as default values in the OPFLOWOptions struct, but that may be above my pay grade.

cameronrutherford: No problem

tests/functionality/sopflow/selfcheck.cpp
bjpalmer: I talked to Bill and got rid of the checks on rank 0. I'm adding a function that checks to see if the error is true anywhere and if it is, then everyone will throw an exception.

tests/functionality/scopflow/selfcheck.cpp
bjpalmer: exit stops the entire program.

tests/functionality/pflow/selfcheck.cpp
bjpalmer: I dialed this back and just use std strings and the c_str() function. According to the documentation that I found, c_str() returns null-terminated c strings so I'm not sure why using pbuf gets rid of the memory error. Sometimes valgrind returns stuff that is impossible to figure out. It doesn't look like changing this modifies the behavior of the test suite.

tests/functionality/scopflow/selfcheck.cpp
bjpalmer: I got rid of pbuf.

tests/functionality/pflow/selfcheck.cpp
cameronrutherford: Yeah I think valgrind will be hard to sift through without a `~mpi` build, but I appreciate the checks.

Again we have other issues with `.c_str` in the code base, so we might revisit this in another PR anyway

tests/functionality/scopflow/selfcheck.cpp
cameronrutherford: Right so maybe we want to use ExaGO's exit/error functionality to ensure a graceful cleanup

tests/functionality/pflow/selfcheck.cpp
bjpalmer: Okay, I'm calling this resolved.

tests/functionality/scopflow/selfcheck.cpp
bjpalmer: I got rid of the exit(0)'s completely. According to Bill, the C++ exceptions should stop the program unless they are caught in a try/catch block by the calling program, so that is probably approximately the behavior we want. If it isn't, the place to fix it would be the ExaGOError functionality.

tests/functionality/pflow/selfcheck.cpp
jaelynlitz: Yeah I think I originally put in the `logging_rank`. I mean I would argue for more configurable code instead of hard coding to 0 but it works either way

jaelynlitz: but yes should be consistent one way or the other

cameronrutherford: I think 0 is a sane default, and if you want we can have this in a private internal ExaGO header.

bjpalmer: I just got rid of logging_rank and replaced it with zero wherever it showed up. It will be pretty easy to add it back in in the future if there is a need for it. Since IO demands in this code are pretty low, just assuming it is 0 is probably a good bet.

tests/functionality/sopflow/selfcheck.cpp
bjpalmer: I modified this slightly. I only do a global check on errors that are not MPI-related. Presumably if something like MPI_Comm_rank fails, then the Allreduce inside the check to see of there is a failure anywhere in the system is going to fail as well.

tpl/pybind11
bjpalmer: I think this happened when did the rebase. I got conflicts with the submodule versions and probably resolved it the wrong way. I think they've been set back to match develop.

tpl/spack
bjpalmer: Same as above

cameronrutherford:
There are 36 commits here and it looks like you pulled in some of @jaelynlitz 's commits. 

I also don't see the correct `#ifdef` logic, even though you commented that it was fixed. Let me know when this is updated and I can re-review

bjpalmer:
I think I've pushed everything at this point and rebased. My version is running okay, which I don't think would happen if those extra endifs are still in place. I believe all the comments have been addressed at this point.

src/opflow/interface/opflowregi.cpp
```diff
@@ -76,10 +76,11 @@ extern PetscErrorCode OPFLOWModelCreate_PBPOLHIOP(OPFLOW);
 
 #if defined(EXAGO_ENABLE_RAJA)
 extern PetscErrorCode OPFLOWModelCreate_PBPOLRAJAHIOP(OPFLOW);
+#endif
```
cameronrutherford: `#ifdef` stil wrong

src/opflow/interface/opflowregi.cpp
cameronrutherford: `#ifdef` stil wrong

src/opflow/interface/opflowregi.cpp
```diff
@@ -117,11 +118,12 @@ PetscErrorCode OPFLOWModelRegisterAll(OPFLOW opflow) {
   ierr = OPFLOWModelRegister(opflow, OPFLOWMODEL_PBPOLRAJAHIOP,
                              OPFLOWModelCreate_PBPOLRAJAHIOP);
   CHKERRQ(ierr);
+#endif
```
cameronrutherford: `#ifdef` stil wrong

src/opflow/interface/opflowregi.cpp
```diff
@@ -117,11 +118,12 @@ PetscErrorCode OPFLOWModelRegisterAll(OPFLOW opflow) {
...
+
 #if defined(EXAGO_ENABLE_HIOP_SPARSE)
   ierr = OPFLOWModelRegister(opflow, OPFLOWMODEL_PBPOLRAJAHIOPSPARSE,
                              OPFLOWModelCreate_PBPOLRAJAHIOPSPARSE);
   CHKERRQ(ierr);
-#endif
```
cameronrutherford: `#ifdef` stil wrong

src/opflow/interface/opflowregi.cpp
```diff
@@ -134,13 +136,13 @@ extern PetscErrorCode OPFLOWSolverCreate_IPOPT(OPFLOW);
 #endif
 #if defined(EXAGO_ENABLE_HIOP)
 extern PetscErrorCode OPFLOWSolverCreate_HIOP(OPFLOW);
+#endif
```
cameronrutherford: `#ifdef` stil wrong

src/opflow/interface/opflowregi.cpp
```diff
@@ -134,13 +136,13 @@ extern PetscErrorCode OPFLOWSolverCreate_IPOPT(OPFLOW);
...
 #if defined(EXAGO_ENABLE_HIOP_SPARSE)
 extern PetscErrorCode OPFLOWSolverCreate_HIOPSPARSE(OPFLOW);
+#endif
```
cameronrutherford: `#ifdef` stil wrong

src/opflow/interface/opflowregi.cpp
```diff
@@ -134,13 +136,13 @@ extern PetscErrorCode OPFLOWSolverCreate_IPOPT(OPFLOW);
...
 #if defined(EXAGO_ENABLE_RAJA)
 extern PetscErrorCode OPFLOWSolverCreate_HIOPSPARSEGPU(OPFLOW);
 #endif
-#endif
```
cameronrutherford: `#ifdef` stil wrong

src/utils/utils.cpp
```diff
@@ -584,12 +584,19 @@ PetscErrorCode ExaGOInitialize(MPI_Comm comm, int *argc, char ***argv,
 PetscErrorCode ExaGOFinalize() {
   PetscFunctionBegin;
   ExaGOLog(EXAGO_LOG_INFO, "Finalizing {} application.", ExaGOCurrentAppName);
+  MPI_Comm comm = MPI_COMM_WORLD;
+  int my_rank;
+  int logging_rank = 0;
```
cameronrutherford: logging rank still here

src/utils/utils.cpp
```diff
@@ -584,12 +584,19 @@ PetscErrorCode ExaGOInitialize(MPI_Comm comm, int *argc, char ***argv,
...
+  auto err = MPI_Comm_rank(comm, &my_rank);
+  if (err)
+    throw ExaGOError("Error getting MPI rank number");
   bool flg;
   ExaGOLogIsUsingLogFile(&flg);
   if (flg) {
     std::string filename;
     ExaGOLogGetLoggerName(filename);
-    ExaGOLog(EXAGO_LOG_INFO, "See logfile {} for output.", filename);
+    if (my_rank == logging_rank)
```
cameronrutherford: and here

tests/functionality/pflow/selfcheck.cpp
```diff
@@ -39,28 +39,38 @@ struct PflowFunctionalityTests
...
   }
 
   void
   ensure_options_are_consistent(toml::value testcase,
                                 toml::value presets = toml::value{}) override {
 
+    int my_rank;
+    auto err = MPI_Comm_rank(comm, &my_rank);
+    if (err != MPI_SUCCESS)
+      throw ExaGOError("Error getting MPI rank number");
```
cameronrutherford: We get rank more than one time here. While not the biggest issue, can you think of a way to only perform this once and to keep this in perhaps the testing initialization routines instead of within these lower level functions?

Since we have an abstract test base, that could also be the place for this

tests/functionality/pflow/selfcheck.cpp
```diff
@@ -108,13 +118,13 @@ struct PflowFunctionalityTests
   void run_test_case(Params &params) override {
     PetscErrorCode ierr;
     PFLOW pflow;
-    int rank;
-
-    MPI_Comm_rank(PETSC_COMM_WORLD, &rank);
+    int my_rank;
+    auto err = MPI_Comm_rank(comm, &my_rank);
```
cameronrutherford: Same here, shouldn't need to repeat this many times

tests/functionality/scopflow/CMakeLists.txt
```diff
@@ -80,8 +80,6 @@ if(EXAGO_INSTALL_TESTS)
     ${RUNCMD}
     $<TARGET_FILE:test_scopflow_functionality>
     ${CMAKE_CURRENT_SOURCE_DIR}/hiop_cont_pbhiop.toml
-    -hiop_mem_space
-    HOST
```
cameronrutherford, line 84: I think this option is helpful?

tests/functionality/scopflow/CMakeLists.txt
```diff
@@ -103,8 +101,6 @@ if(EXAGO_INSTALL_TESTS)
     3
     $<TARGET_FILE:test_scopflow_functionality>
     ${CMAKE_CURRENT_SOURCE_DIR}/hiop_cont_mpi_pbhiop.toml
-    -hiop_mem_space
-    HOST
```
cameronrutherford, line 107: This also seems useful

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -104,17 +108,28 @@ struct ScopflowFunctionalityTestParameters {
...
 
   using Params = ScopflowFunctionalityTestParameters;
+  MPI_Comm comm;
+  int nprocs;
+
+  ScopflowFunctionalityTests(std::string testsuite_filename, MPI_Comm comm,
+                             int logging_verbosity = EXAGO_LOG_INFO)
+      : FunctionalityTestContext(testsuite_filename, logging_verbosity),
+        comm{comm} {
+    auto err = MPI_Comm_size(comm, &nprocs);
```
cameronrutherford: Same here - we don't want to repeat ourselves if possible

tests/functionality/scopflow/selfcheck.cpp
```diff
@@ -429,7 +444,14 @@ int main(int argc, char **argv) {
   ExaGOCheckError(ierr);
   ExaGOLog(EXAGO_LOG_INFO, "{}", "Creating SCOPFlow Functionality Test");
 
-  ScopflowFunctionalityTests test{std::string(argv[1])};
+  int my_rank;
+  auto err = MPI_Comm_rank(comm, &my_rank);
```
cameronrutherford, line 442: Same here

tests/functionality/sopflow/selfcheck.cpp
```diff
@@ -112,17 +112,33 @@ struct SopflowFunctionalityTestParameters {
...
 
   using Params = SopflowFunctionalityTestParameters;
+  MPI_Comm comm;
+  int nprocs;
+
+  SopflowFunctionalityTests(std::string testsuite_filename, MPI_Comm comm,
+                            int logging_verbosity = EXAGO_LOG_INFO)
+      : FunctionalityTestContext(testsuite_filename, logging_verbosity),
+        comm{comm} {
+
```
cameronrutherford, line 121: Same here

cameronrutherford (changes requested):

cameronrutherford:
Maybe I am just seeing different things from what you are seeing, but ifdefs are still wrong. Left some other comments for potential improvements.

tests/functionality/scopflow/CMakeLists.txt
bjpalmer: It's covered in the .toml files.

tests/functionality/scopflow/CMakeLists.txt
bjpalmer: Also covered in the .toml file.

src/opflow/interface/opflowregi.cpp
bjpalmer: I guess I'm not seeing it. How is it wrong?

src/opflow/interface/opflowregi.cpp
bjpalmer: Check out the actual source file instead of the diffs https://github.com/pnnl/ExaGO/blob/func-test-mpi/src/opflow/interface/opflowregi.cpp. It looks okay to me.

tests/functionality/scopflow/selfcheck.cpp
bjpalmer: Not sure what you are referring to here. What is repeating?

src/opflow/interface/opflowregi.cpp
cameronrutherford: Let me check the source code directly in your branch...

src/opflow/interface/opflowregi.cpp
cameronrutherford: Okay I checked the source files and my comments stand.

@abhyshr and @wperkins can clarify here, but taking our sparse and sparse_gpu models here as an example.

If we build with sparse, we need to enable the sparse model. Then, within that conditional, we also need to check for RAJA to determine if we support sparse_gpu.

The key here is the conditionals are nested, whereas your changes has moved these conditionals to be sequential. Your changes modify behaviour

tests/functionality/scopflow/selfcheck.cpp
cameronrutherford: The call to get rank and the associated error checking is repeated.

Maybe not worth the effort, but I was thinking there could be a quick way to move this to be common code across all tests, and not code repeated in each tests source files.

src/opflow/interface/opflowregi.cpp
bjpalmer: Okay, catch your drift now. I've modified it to match the develop branch and will check it in if the test suite runs.

tests/functionality/scopflow/selfcheck.cpp
bjpalmer: I can add a utility function that does this. It would be slicker if the FunctionalityTestContext was MPI-aware, but that would be a bigger lift. Although...
Now that a take a quick look, it wouldn't be that big a lift. I'll see what I can do.

bjpalmer: Okay, I've made the changes and rebased against develop. The checks are still in main, but I got rid of them everywhere else. If this passes then I'm good.

tests/functionality/sopflow/selfcheck.cpp
```diff
@@ -431,7 +438,15 @@ int main(int argc, char **argv) {
   ExaGOCheckError(ierr);
   ExaGOLog(EXAGO_LOG_INFO, "{}", "Creating SOPFLOW Functionality Test");
 
-  SopflowFunctionalityTests test{std::string(argv[1])};
+  int my_rank;
+  auto err = MPI_Comm_rank(comm, &my_rank);
```
cameronrutherford, line 442: Still some refactoring to do in other application tests to refactor out this rank stuff.

src/utils/utils.cpp
cameronrutherford: Still logging rank here.

src/utils/utils.cpp
```diff
@@ -584,12 +584,19 @@ PetscErrorCode ExaGOInitialize(MPI_Comm comm, int *argc, char ***argv,
 PetscErrorCode ExaGOFinalize() {
   PetscFunctionBegin;
   ExaGOLog(EXAGO_LOG_INFO, "Finalizing {} application.", ExaGOCurrentAppName);
+  MPI_Comm comm = MPI_COMM_WORLD;
+  int my_rank;
+  int logging_rank = 0;
+  auto err = MPI_Comm_rank(comm, &my_rank);
```
cameronrutherford, line 589: Still remnant rank stuff to refactor here.

cameronrutherford (commented):
Looking awesome!! Just a few more small fixes

tests/functionality/sopflow/selfcheck.cpp
bjpalmer: This occurs outside of FunctionalityTestContext which is initialized later in the code. I'd just leave this one (there is similar code in the `main` routines in all the other `selfcheck.cpp` codes).

src/utils/utils.cpp
bjpalmer: Again, this falls outside of the FunctionalityTestContext. The better way to handle this would have been to create an  environment object, initialize rank and nprocs inside that and then just access the values whenever you need them. Since these are just function calls, that doesn't work. I'd just leave it. If you want, you could get rid of the actual check on the error since if the code made it this far it is extremely unlikely that MPI_Comm_rank will fail.

bjpalmer:
The logging_rank in utils.cpp is gone.

cameronrutherford:
> The logging_rank in utils.cpp is gone.

https://github.com/pnnl/ExaGO/blob/func-test-mpi/src/utils/utils.cpp#L589 still looks there to me

bjpalmer:
Forgot to commit. It should be gone now.

cameronrutherford:
Your tests were failing because I killed GitLab CI for a bit there in #85. My bad - I have kicked off fresh pipelines and once tests pass we can merge.

Thanks again for this bruce - makes our debugging much quicker

cameronrutherford (approved):

PR #29, opened 2023-10-04
Draft: upgrade to gcc11

jaelynlitz:


buildsystem/spack/deception/env.sh
```diff
@@ -8,8 +8,8 @@ module load python/miniconda3.9
 source /share/apps/python/miniconda3.9/etc/profile.d/conda.sh
 
 # Load system modules
-module load gcc/9.1.0
-module load cuda/11.4
+module load gcc/11.2.0
+module load cuda/11.8
```
cameronrutherford, line 12: Hopefully we don't need specifically 11.3.0... If we need to switch to that, seems like we would also have to downgrade to CUDA 11.7...

PR #41, opened 2023-10-12
Add Spack CPU build with `exago+hiop+raja~ipopt ^hiop+raja~sparse`

### LINKED ISSUE #40: `exago+hiop+raja~ipopt ^hiop+raja~sparse` fails to build
https://github.com/xsdk-project/xsdk-issues/issues/226 - thanks to @balos1 for catching this.

Copying from the xSDK issue:

> @abhyshr and @wperkins this error is occuring with a build of exago+mpi+python+ipopt~rocm~cuda. Key here is ~ipopt means https://github.com/pnnl/ExaGO/blob/828db06af5e0345b641b6d03ef7e2456d13469ea/src/opflow/interface/opflowregi.cpp#L79 and https://github.com/pnnl/ExaGO/blob/828db06af5e0345b641b6d03ef7e2456d13469ea/src/opflow/model/power_bal_hiop/pbpolrajahiopsparse.cpp#L3 end up being guarded by different compiler macros.

cameronrutherford:
With the additional GitHub action we should be able to reproduce issue identified in xSDK testing, so we can add a fix that is verifiable in local pipelines.

Closes:
- #40 

xSDK issues:
- https://github.com/xsdk-project/xsdk-issues/issues/231
- https://github.com/xsdk-project/xsdk-issues/issues/226

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -11,25 +11,29 @@ jobs:
     container: spack/ubuntu-focal:latest
     strategy:
       matrix:
+        # Minimal Build(s)
+        # Need S3 mirror to have these builds speedup
         spack_spec:
-          # Minimal Build(s)
           # See #39 - ~python~mpi causes issues
         # - exago@develop~mpi~ipopt~hiop~python~raja
+          # See #44 - +mpi~python should fail if no python in system,
+          #                       but the runner happens to have one...
           - exago@develop+mpi~ipopt+hiop~python~raja ^openmpi
           # See #18 - +hiop~mpi causes issues
         # - exago@develop~mpi~ipopt+hiop~python+raja
           # See #16 - +python~mpi causes issues
         # - exago@develop~mpi~ipopt+hiop+python~raja
           - exago@develop+mpi~ipopt+hiop+python~raja ^openmpi
-        # Builds with hiop and without Ipopt are useful for edge cases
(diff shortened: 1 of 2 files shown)


.github/workflows/spack_cpu_build.yaml
```diff
@@ -21,6 +21,8 @@ jobs:
           # See #16 - +python~mpi causes issues
         # - exago@develop~mpi~ipopt+hiop+python~raja
           - exago@develop+mpi~ipopt+hiop+python~raja ^openmpi
+        # Builds with hiop sparse and without Ipopt are useful for edge cases
+          - exago@develop~mpi~ipopt+hiop~python+raja ^hiop+raja+sparse
```
cameronrutherford, line 26:
```suggestion
        # Builds with hiop and without Ipopt are useful for edge cases
          - exago@develop~mpi~ipopt+hiop~python+raja ^hiop+raja~sparse

```

cameronrutherford:
Added `hiop+sparse` initially, which both doesn't work in GitHub actions, and doesn't repro the issue... Re-trying with the correct build

cameronrutherford:
I re-produced #18... Going to add build with MPI and try to repro original issue #40

cameronrutherford:
Successfully reproduced the issue.  Just need to write the fix

cameronrutherford:
@wperkins and @abhyshr please review and make sure I am `#ifdef` guarding the right things here. If GitHub actions (and other pipelines) pass then we are good to merge and close the xSDK bug.

wperkins (approved):
This looks fine to me.

PR #47, opened 2023-10-13
Spack CPU build `exago~mpi`

### LINKED ISSUE #16: `exago+python~mpi` fails to build
https://github.com/pnnl/ExaGO/actions/runs/6301983110/job/17108216891?pr=15

mpi4py is included regardless of if MPI is configured. Should be an easy fix, and a GitHub action pipeline should be configured so ensure this sort of failure is prevented in the future.

See #15

### LINKED ISSUE #39: `exago~mpi~python` fails to build
Similar to #16 but additionally `~python`

### LINKED ISSUE #48: `exago~mpi` causes build issues
This is a collection of smaller individual build failures, and is capturing the higher level issue:

- [ ] #16 
  - This started as a relatively basic CMake error resolve in #47, but still won't work until `~mpi` works
- [ ] #18
  - This might be a HiOp issue, or an ExaGO error, or both. More importantly this doesn't get caught because it isn't tested
- [ ] #39 
  - Perhaps just because no-one builds this configuration, this never seems to work well

cameronrutherford:
This PR will first reproduce the failures (or lack thereof), and then we can add and merge in fixes.

Closes:
- [x] #16
- [x] #39

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -14,17 +14,14 @@ jobs:
         # Minimal Build(s)
         # Need S3 mirror to have these builds speedup
         spack_spec:
-          # See #39 - ~python~mpi causes issues
-        # - exago@develop~mpi~ipopt~hiop~python~raja
-          # See #44 - +mpi~python should fail if no python in system,
-          #                       but the runner happens to have one...
-          - exago@develop+mpi~ipopt+hiop~python~raja ^openmpi
+          # See #39 - minimal build useful for sanity
+          - exago@develop~mpi~ipopt~hiop~python~raja
+          # See #16 / #44 - +python~mpi catches mpi4py / python config
+          - exago@develop~mpi~ipopt~hiop+python~raja
           # See #18 - +hiop~mpi causes issues
         # - exago@develop~mpi~ipopt+hiop~python+raja
-          # See #16 - +python~mpi causes issues
-        # - exago@develop~mpi~ipopt+hiop+python~raja
           - exago@develop+mpi~ipopt+hiop+python~raja ^openmpi
-          # See #40 - +hiop+raja~ipopt ^hiop~sparse is useful for edge cases
+          # See #40 - +hiop+raja~ipopt ^hiop~sparse is useful for edge cases in xSDK
           - exago@develop+mpi~ipopt+hiop~python+raja ^openmpi ^hiop+raja~sparse 
 
     name: Build ExaGO with Spack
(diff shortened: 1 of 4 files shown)


.github/workflows/spack_cpu_build.yaml
```diff
@@ -14,17 +14,14 @@ jobs:
...
         spack_spec:
-          # See #39 - ~python~mpi causes issues
-        # - exago@develop~mpi~ipopt~hiop~python~raja
-          # See #44 - +mpi~python should fail if no python in system,
-          #                       but the runner happens to have one...
-          - exago@develop+mpi~ipopt+hiop~python~raja ^openmpi
+          # See #39 - minimal build useful for sanity
+          - exago@develop~mpi~ipopt~hiop~python~raja
+          # See #16 / #44 - +python~mpi catches mpi4py / python config
+          - exago@develop~mpi~ipopt~hiop+python~raja
```
cameronrutherford, line 20: Issue in `+mpi~python` should be equivalent to a build with `~mpi~python` as far as I can tell. Either way I think this is sufficient and we need to ensure these builds pass in future.

.github/workflows/spack_cpu_build.yaml
```diff
@@ -41,8 +38,10 @@ jobs:
           SPACK_SPEC: ${{ matrix.spack_spec }}
         run: |
           ls && pwd
-          . ./tpl/spack/share/spack/setup-env.sh
+          . /opt/spack/share/spack/setup-env.sh
```
cameronrutherford, line 41: This is more sustainable, and can be swapped out when we want to use ExaGO's spack. 

It might just make more sense to have a `package.py` stored in the repo for ExaGO as that is the main thing that is regularly modified instead of relying on submodules. Still would be nice to pin a spack version and support that workflow, but making it an optional clone would be best.

cameronrutherford:
Marking as draft for now.

While this PR would in theory solve all our build issues, the real problem (captured in #48), is that we don't build/test with MPI.

For now, this will have to wait for 1.6.0

.github/workflows/spack_cpu_build.yaml
jaelynlitz: The spack repo is is incredibly active, I would be hesitant to pin a version of spack because that's just another package we would could get hung up on a "blessed" version or forget to upgrade that one clone...

cameronrutherford: To clarify, for 100% reproducible spack builds we need to pin the version in certain situations (such as profiling runs, or key performance runs).

For CI and automated builds, I do think that just using spack's develop like we do with this base image is the best way to go.

Is that in line with what you were thinking?

cameronrutherford:
There are some valuable CMake changes that should be added to ExaGO here @abhyshr. You should also probably build and test without MPI...

PR #49, opened 2023-10-17
Only check for Python if it is enabled

### LINKED ISSUE #44: `exago+mpi~python` fails to build
https://gitlab.com/xsdk-project/spack-xsdk/-/jobs/5273380918 - CMake still tries to look for Python and fails to find it

From https://github.com/xsdk-project/xsdk-issues/issues/224

Closely related to #39, but `+mpi` instead

cameronrutherford:
Since #48 was identified, I realize that fixing builds `+python~mpi` #16 would entail closing #48, so this is a more modest PR just to resolve #44 

We have a spack pipeline that should catch this error in theory, but it is picking up the system python, so only xSDK pipelines seem to be reproducers at this point. I don't think it is worth the time to develop pipelines to repro this since it is a small fix and we have a release coming.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,8 +406,8 @@ else()
   endif()
 endif()
 
-include(ExaGOCheckPython)
 if(EXAGO_ENABLE_PYTHON)
+  include(ExaGOCheckPython)
   # Build pybind11 target manually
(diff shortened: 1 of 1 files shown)


wperkins (approved):
Yes, this is the logical thing to do.

PR #51, opened 2023-10-17
Fix Ascent Logger Test

### LINKED ISSUE #42: `UNIT_TESTS_LOGGER` fails on Ascent
Introduced in #20 

https://code.ornl.gov/ecpcitest/exasgd/exago/-/jobs/2072667

jaelynlitz:
closes #42 

Somehow even after deleting the spack built mpi, the something in `dependencies.sh` [is still loading it ](https://code.ornl.gov/ecpcitest/exasgd/exago/-/jobs/2083092#L64)so I forced loading the system module afterwards and that fixes it. I am aware that this is a little janky.

### DIFF
--- a/buildsystem/gcc-cuda/ascentVariables.sh
+++ b/buildsystem/gcc-cuda/ascentVariables.sh
@@ -8,4 +8,3 @@ source $SRCDIR/buildsystem/spack/$MY_CLUSTER/modules/dependencies.sh
 
 # Shared system configuration
 source $SRCDIR/buildsystem/gcc-cuda/ascent/base.sh
-
--- a/buildsystem/gcc-cuda/ascentVariables.sh
+++ b/buildsystem/gcc-cuda/ascentVariables.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
 export SRCDIR=${SRCDIR:-$PWD}
-
-# Shared system configuration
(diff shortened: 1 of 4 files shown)


buildsystem/gcc-cuda/ascentVariables.sh
```diff
@@ -7,3 +7,5 @@ source $SRCDIR/buildsystem/gcc-cuda/ascent/base.sh
 
 # Spack modules
 source $SRCDIR/buildsystem/spack/$MY_CLUSTER/modules/dependencies.sh
+
+module load spectrum-mpi/10.4.0.3-20210112
```
cameronrutherford, line 11:
```suggestion

# Spack modules
source $SRCDIR/buildsystem/spack/$MY_CLUSTER/modules/dependencies.sh
```

IMO what we could do is:
- Remove `module purge` equivalents from scripts, which might nuke other friendly things that are loaded by default
- Load spack modules, then load `base.sh`. That way we can keep all custom and weird fixes to the environment in there, and also ensure it is run last

cameronrutherford (approved):
Thanks for fixing this!

cameronrutherford (approved):

buildsystem/gcc-cuda/ascentVariables.sh
cameronrutherford: I was suggesting extending this to all platforms, but this is great for now. Thanks!

Issue #35, opened 2023-10-11
Jupyter Notebook Support / Python API additions


cameronrutherford:
For future ExaGO documentation, having `.ipynb` files available for reference is important. In addition, @abhyshr has discussed the desire to post-process ExaGO data in Python, which might require API changes.

To resolve the issue:
- [x] Create basic Jupyter notebook without MPI running OPFLOW example
- [x] Write OPFLOW output to disk with `saveSolution`, and then read that back into Python staying in JSON (maybe convert to dict...)
- [ ] Add Python function that returns a dictionary of the OPFLOW output without needing to read/write to disk

These activities should be considered in the context of #32, as Jupyter notebooks are easy to render and document alongside Quarto.

cameronrutherford:
https://zhauniarovich.com/post/2020/2020-06-clearing-jupyter-output/ - we should render #32 Quarto docs for our project in GitHub actions, and then use our exago spack version to generate our docs.

cameronrutherford:
Would be great to add a notebook/configuration to compare linear solvers using MA57, ReSolve and ReSolve GPU as backend to Ipopt. cc @pelesh

cameronrutherford:
Is it possible to run mpi4py within Jupyter? https://github.com/mpi4py/mpi4py/discussions/294

If so, lets get that working for a demo!

If not, need to resolve #16

cameronrutherford:
@abhyshr  and @jaelynlitz can we close this out? Maybe we can make a new issue that better summarizes remaining work?

#85 made me realize we have closed out Jupyter and base example side of things, but it might be easier to just consolidate and move python API side of things to new issue since different end users might have different needs and ways to connect there data outputs...

abhyshr:
Good to close. We already have Jupyter notebooks in develop now.

Issue #52, opened 2023-10-19
`macOS` GitHub Spack Action


cameronrutherford:
Should be possible according to pnnl GitHub admins, so would be good to buld (and even test!?) that configuration

PR #64, opened 2023-11-05
devcontainer support with tutorial jupyter notebook

cameronrutherford:
**Merge request type**
- [x] New feature

**Relates to**
- [x] Web docs
- [x] Other

**This MR updates**
- [x] Web docs
- [x] Manual
- [x] Other

**Summary**

Pre-cursor to #35 to support working locally on MacOS and other systems. Pre-requisites are:
- Docker daemon installed/working
- VSCode installed w/ remote-development extensions enabled
- CoinHSL 2019 tarball downloaded and executable in the ExaGO source dir with correct name


docs/devcontainer/tutorial.ipynb
cameronrutherford, line 1: TODO

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+# What we want to install and how we want to install it
+# is specified in a manifest file (spack.yaml)
+RUN mkdir /opt/spack-environment \
+&&  (echo spack: \
+&&   echo '  specs:' \
+&&   echo '  - exago@1.6.0+ipopt+python+mpi' \
+&&   echo '  - petsc~fortran~hdf5~hypre+metis' \
+&&   echo '  - ipopt@3.12.10+coinhsl~mumps+metis' \
+&&   echo '  - coinhsl@2019.05.21+blas' \
+&&   echo '  - py-jupyterlab' \
```
cameronrutherford, line 18: 2336.9s was how long it took me to build on my laptop, which installed 171 packages.  

cc @paulrigor and @Jayapreethi this is the sort of use case we need to support and speed up for users if possible

cameronrutherford:
These changes will never affect CI, so I think we are good to merge after some careful review.

Still several steps to finally close out #35, but this is the MVP

.devcontainer/Dockerfile
cameronrutherford: @eugeneswalker maybe E4S can help here?

We can't use CoinHSL, but it would at least be nice to cache some of these packages to speed up people's devcontainer builds when they want to try things out for themselves. 

SInce E4S support ends after ECP, we need a sustainable solution that ExaGO can adopt independently after project ends at low cost.

cc @pelesh as we are almost ready with viable CoinHSL alternative, which would allow us to provide true rolling ExaGO releases, squashing 171 packages into needing <10...

cameronrutherford:
If we want to add a build without Ipopt, we can also support launching ExaGO directly in GitHub codespaces...

.devcontainer/Dockerfile
eugeneswalker: > @eugeneswalker maybe E4S can help here?

Yes, we can help by putting as many of these packages into a cache as possible. This container is intended to run on intel or arm macOS, or both and more?

cameronrutherford: With dev-containers and container runtimes, I was under the impression the we would just need ubuntu x86_64 binaries, and then those would effectively be cross platform?

That being said, we also want to support users running CPU only builds on their laptops with all the architectures that you listed, so if MacOS is within reach I would love that too.

#52 is working towards this in that at least having GitHub actions for Mac would be good, but ideally supporting this fully would be what we do

jaelynlitz (changes requested):
Please add documentation on how to build and run the container

jaelynlitz (approved):

PR #67, opened 2023-11-08
Add function to demo nb to convert CSV output file to JSON

jaelynlitz:
**Merge request type**
- [X] New feature
- [ ] Resolves bug
- [ ] Documentation
- [ ] Other

**Relates to**
- [ ] OPFLOW
- [ ] SOPFLOW
- [ ] SCOPFLOW
- [ ] TCOPFLOW
- [ ] CMake build system
- [ ] Spack configuration
- [ ] Manual
- [ ] Web docs
- [X] Other (Python demo)

**This MR updates**
- [ ] Header files
- [ ] Source code
- [ ] CMake build system
- [ ] Spack configuration
- [ ] Web docs
- [ ] Manual
- [X] Other (Python demo)

**Summary**

In the tutorial notebook, convert an opflow output CSV file to a JSON dictionary

**Linked Issue(s)**
Partially addresses #35


cameronrutherford:
This is the first iteration of support. More important feature is Python api of `getSolution` that will return the solution in-place.

We should also make sure this approach and the API `getSolution` return the same data type/structure somehow...

jaelynlitz:
> This is the first iteration of support. More important feature is Python api of `getSolution` that will return the solution in-place.
> 
> We should also make sure this approach and the API `getSolution` return the same data type/structure somehow...

Right, this is bullet 2 in #35  - with the goal to get things working for Shri by Monday. There are some things rn that I imagine are hard-coded to the OPFLOW output that I'm not sure how they'll differ for SCOPFLOW for example.

jaelynlitz:
rebuilding the container to double check that everything still works then ready to merge if you agree @abhyshr

docs/devcontainer/tutorial.ipynb
```diff
@@ -133,65 +146,166 @@
...
+      "data converted from ./tmp.output to ./data.json\n"
+     ]
+    }
+   ],
+   "source": [
+    "import csv \n",
```
abhyshr, line 214: Can we put this source in a separate python file and then import the function. I think this source will be useful beyond the notebook.

abhyshr:
Can you please put the CSV->JSON conversion code in a separate file? Also, one can directly load the JSON output file created by OPFLOW using the JSON module in python. That's another alternative for users.

abhyshr:
Good to merge when the seperate source file is created.

abhyshr (approved):

docs/devcontainer/tutorial.ipynb
cameronrutherford: The idea is just to show users an example of how to roll their own, and then we can truly support this from the python wrapper in another PR with C++ code that generates an actual python API.

cameronrutherford:
> Good to merge when the seperate source file is created.

@abhyshr I think we should merge this, and then in another PR create a separate notebook talking about I/O, and use this as the starting point. 

Would be good to also do a small benchmark of the different approaches in this new notebook, and provide suggestions to users. That way we can also encourage future ExaGO users to contribute different/better examples as the project evolves.

docs/devcontainer/sample_opflow.output
cameronrutherford, line 1: Shouldn't make it a habit to add too many of these, but I think this is helpful so users don't have to re-generate.

cameronrutherford (approved):

Issue #43, opened 2023-10-12
OPFLOW HiOp functionality tests failing with `hiop@1.0.0`


cameronrutherford:
From GitLab, this was happening on Newell, Deception and Incline - https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/issues/469

Now is is also happening on Ascent in #20 - https://code.ornl.gov/ecpcitest/exasgd/exago/-/jobs/2072667

These tests are skipped through a labels in `tests/functionality/opflow/CMakeLists.txt`

jaelynlitz:
now fixed on Newell in #84

jaelynlitz:
now fixed on Deception in #84 https://gitlab.pnnl.gov/exasgd/frameworks/exago-github-mirror/-/jobs/138063

cameronrutherford:
Incline and Ascent next ;)

What about similar tests in #3?

jaelynlitz:
enabling to find out...

jaelynlitz:
Incline's sole OPFLOW failure is a seg fault on test `FUNCTIONALITY_TEST_OPFLOW_RAJAHIOP_SPARSE_GPU_TOML_TESTSUITE` 

https://gitlab.pnnl.gov/exasgd/frameworks/exago-github-mirror/-/jobs/138118#L108

cameronrutherford:
> Incline's sole OPFLOW failure is a seg fault on test `FUNCTIONALITY_TEST_OPFLOW_RAJAHIOP_SPARSE_GPU_TOML_TESTSUITE`
> 
> https://gitlab.pnnl.gov/exasgd/frameworks/exago-github-mirror/-/jobs/138118#L108

Sounds like you can disable that test on incline, and create new PR since test failure is isolated. Would be worth checking for identical behavior on Crusher (using #34 or otherwise), and making sure to use older ROCm version to match incline (unlike #89)

PR #68, opened 2023-11-09
Use #if defined instead of #if and turn off -Werror for debug builds

abhyshr:
`#if` does not work correctly on Mac so switching to `#if defined`. Noticed this behavior for
`#if (EXAGO_ENABLE_IPOPT || EXAGO_ENABLE_HIOP)` so replaced it with
`#if  defined EXAGO_ENABLE_IPOPT || defined EXAGO_ENABLE_HIOP`.

Also removed the `-Werror` flag for debug builds. `sprintf` is depecreated for clang and so `-Werror` flag results in zillions of errors. Note that the error is thrown not only for `sprintf` in ExaGO code, but also if it is any dependency library. Need to tackle this through a separate issue so turning off the `-Werror` for now.

### DIFF
--- a/src/utils/CMakeLists.txt
+++ b/src/utils/CMakeLists.txt
@@ -33,10 +33,10 @@ exago_add_library(
 # works -Wpedantic for CUDA compile options added a significant amount of new
 # errors, so leaving out for now
 set(CUDA_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Werror,-Wall,-Wextra>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Wall,-Wextra>"
 )
 set(CXX_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Werror;-Wall;-Wextra;-Wpedantic>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Wall;-Wextra;-Wpedantic>"
 )
 if(EXAGO_BUILD_SHARED)
   target_compile_options(
--- a/src/utils/CMakeLists.txt
+++ b/src/utils/CMakeLists.txt
@@ -30,9 +30,9 @@ exago_add_library(
 )
 
 # Going to have this target supply/propogate compiler flags and see if that
-# works -Wpedantic for CUDA compile options added a significant amount of new
-# errors, so leaving out for now -Werror is problematic on MacOS
-if(APPLE)
+# works. -Wpedantic for CUDA compile options added a significant amount of new
+# errors, so leaving out for now. -Werror is problematic on MacOS
+if("${CMAKE_CXX_COMPILER_ID}" MATCHES "AppleClang")
   set(CUDA_COMPILE_OPTIONS
       "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Wall,-Wextra>"
   )
--- a/src/utils/CMakeLists.txt
+++ b/src/utils/CMakeLists.txt
@@ -31,13 +31,23 @@ exago_add_library(
 
 # Going to have this target supply/propogate compiler flags and see if that
 # works -Wpedantic for CUDA compile options added a significant amount of new
-# errors, so leaving out for now
-set(CUDA_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Wall,-Wextra>"
-)
-set(CXX_COMPILE_OPTIONS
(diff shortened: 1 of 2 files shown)


src/utils/CMakeLists.txt
```diff
@@ -33,10 +33,10 @@ exago_add_library(
 # works -Wpedantic for CUDA compile options added a significant amount of new
 # errors, so leaving out for now
 set(CUDA_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Werror,-Wall,-Wextra>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Wall,-Wextra>"
 )
 set(CXX_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Werror;-Wall;-Wextra;-Wpedantic>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Wall;-Wextra;-Wpedantic>"
```
cameronrutherford: Turns out it's really easy to add platform specific CMake in cases like this. https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Write-Platform-Checks#platform-variables

You can just do:

```CMake
if(APPLE)

endif()
```

cameronrutherford (changes requested):
I was struggling to add the suggested changes in the GitHub IDE, but let's add the MacOS stuff into a conditional, and set everything else to the current setting.

cameronrutherford:
Also this is a +1 for #52

cameronrutherford:
I will PR into this with CMake suggestion. Still not sure why tests are failing on Newell + Ascent. Noteably two different tests are failing.

cameronrutherford (approved):

cameronrutherford (dismissed):

bjpalmer (dismissed):

cameronrutherford:
Newell and Ascent are still failing so if someone can take a look into those failures then I am happy to merge

src/utils/CMakeLists.txt
wperkins, line 1: Should there be some testing of the compiler here?  What compiler is assumed here, Apple CLang? Does the same thing happen w/ GNU.  I guess I'm asking is this a (host) compiler, (gpu) compiler, or system issue?  And, I only ask this because I don't like the Apple CLang compiler.

cameronrutherford: That's a good question. @abhyshr should we just be doing this with apple clang, or all Apple devices?

cameronrutherford:
Since this seems to have gotten lost in the comments - tests are failing on Newell and Ascent. Please fix

src/utils/CMakeLists.txt
abhyshr: I've seen this on my Mac using Clang. Not tested with any other compiler. The issue is that #if does not work with Clang so one needs to use #if defined instead. This is a fix that I thought would work, but maybe it is not (?). What are the other solutions?

abhyshr: Looking at the PETSc code, I see that they use`#if defined` convention. Since PETSc has builds across many more platforms than we do, I think following their convention would be better for us as well, i.e., using `#if defined` would be better than just `#if`

wperkins: Sorry, I should have been more specific. I was commenting on abandoning  `-Werror` on `APPLE` regardless of the compiler used. Fixing `#if` behavior is fine w/ me.

cameronrutherford: It's relatively easy to limit the compiler flag based on only Apple Clang, and it sounds like that's what we should be doing?

I can create a PR for this if that's the consensus

abhyshr: Without the `-Werror` stuff, is this PR good to be merged?

cameronrutherford: Tests are failing on Newell and Ascent, so no

cameronrutherford:
> Since this seems to have gotten lost in the comments - tests are failing on Newell and Ascent. Please fix

@abhyshr

src/utils/CMakeLists.txt
wperkins: > It's relatively easy to limit the compiler flag based on only Apple Clang, and it sounds like that's what we should be doing?
> 
> I can create a PR for this if that's the consensus

I would vote for this.

cameronrutherford: Happy to add the apple clang conditional, but we still have test failures

bjpalmer:
I just ran this branch on deception and tests 20 and 21 are passing. These were the ones that you were reporting as failing, correct? I'm only seeing failures for 17,18,19.

cameronrutherford:
> I just ran this branch on deception and tests 20 and 21 are passing. These were the ones that you were reporting as failing, correct? I'm only seeing failures for 17,18,19.

Remember we still are skipping some tests in CI per #43 and #3, which @jaelynlitz is also helping resolve. I think 17,18,19 are all skipped, and so you should see Deception tests passing as is reflected in PR status.

You will only see test failures on Ascent and Newell, as is reflected in GitHub commit status.

cameronrutherford (approved):

PR #70, opened 2023-11-09
Fix ignoring contingencies when the from bus number is greater than than to bus number

### LINKED ISSUE #26: [SCOPFLOW] branch contingencies ignored based on to-from bus indices
ExaGO/src/ps/psreaddata.cpp : 899 there is a section of code that swaps branch indices so that the smaller bus index becomes from and the larger to. If contingencies do not match this convention they are ignored by ExaGO, even if the indices are as they appear in the original case file.

abhyshr:
Fixes #26 

When the from bus number is greater than the to bus number for a branch, the PS branch object stores it in the reverse order. This is done so that the from bus number is always less than to bus number. The reason for doing this is IPOPT complains for Hessian if this is not done. However, this was causing the contingencies with from bus > to bus to be ignored.

Fixed this issue by having the `PSGetLine` routine check for both the combinations (from bus -- to bus, and to bus -- from bus)

### DIFF
--- a/src/utils/CMakeLists.txt
+++ b/src/utils/CMakeLists.txt
@@ -33,10 +33,10 @@ exago_add_library(
 # works -Wpedantic for CUDA compile options added a significant amount of new
 # errors, so leaving out for now
 set(CUDA_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Werror,-Wall,-Wextra>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Wall,-Wextra>"
 )
 set(CXX_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Werror;-Wall;-Wextra;-Wpedantic>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Wall;-Wextra;-Wpedantic>"
 )
 if(EXAGO_BUILD_SHARED)
   target_compile_options(
--- a/src/utils/CMakeLists.txt
+++ b/src/utils/CMakeLists.txt
@@ -33,10 +33,10 @@ exago_add_library(
 # works -Wpedantic for CUDA compile options added a significant amount of new
 # errors, so leaving out for now
 set(CUDA_COMPILE_OPTIONS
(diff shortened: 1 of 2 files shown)


abhyshr:
This change is most likely going to cause test failures for SCOPFLOW if any contingencies were being ignored.

abhyshr:
This is ready for merging.

src/utils/CMakeLists.txt
```diff
@@ -33,10 +33,10 @@ exago_add_library(
 # works -Wpedantic for CUDA compile options added a significant amount of new
 # errors, so leaving out for now
 set(CUDA_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Werror,-Wall,-Wextra>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:DEBUG>>:SHELL:--compiler-options -Wall,-Wextra>"
 )
 set(CXX_COMPILE_OPTIONS
-    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Werror;-Wall;-Wextra;-Wpedantic>"
+    "$<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CONFIG:DEBUG>>:-Wall;-Wextra;-Wpedantic>"
 )
```
cameronrutherford: This was originally added by me to repro xSDK CI failures that we see with other compilers. I would love to keep them in as many builds as possible to ensure that our builds are truly cross-platform.

I can limit this to only a few CI platforms if that helps, but I don't see how this would solve any issues by removing.

cameronrutherford (changes requested):
I don't understand why you need to remove `-Werror` from our Debug builds, but your ps changes are sound

src/utils/CMakeLists.txt
abhyshr: I agree that we want to keep this check for all platforms. I think I'll need to create another PR to fix compile errors with Mac.

cameronrutherford: Can I just PR into this with the fix? It's like 5 lines total

cameronrutherford: Reverting these changes in favor of #68 is my suggestion here

cameronrutherford: On the other hand we can just remove these changes here and merge atomic changes one at a time

abhyshr: Yup, please revert this change and merge the branch if the tests pass.

cameronrutherford (approved):

abhyshr:
This is good to merge.

abhyshr:
Closes #26

PR #87, opened 2023-11-22
Update CMakeLists.txt for AppleClang

cameronrutherford:
Closes thread in #68


cameronrutherford:
@abhyshr same tests that are failing in your PR are failing here. I think this is good to merge

PR #105, opened 2023-12-14
Change path padded length in Spack containers

cameronrutherford:
**Merge request type**
- [x] Resolves bug

**Relates to**
- [x] Spack configuration

**This MR updates**
- [x] Spack configuration

**Summary**

Having issues with paths not existing in #4, so attempting to reduce the padded path length to make things easier.

cc @Jayapreethi and @paulrigor

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -118,7 +118,7 @@ jobs:
             config:
               install_tree:
                 root: /opt/spack
-                padded_length: 128
+                padded_length: 64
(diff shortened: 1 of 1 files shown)


.github/workflows/spack_cpu_build.yaml
```diff
@@ -118,7 +118,7 @@ jobs:
             config:
               install_tree:
                 root: /opt/spack
```
cameronrutherford, line 120: @Jayapreethi @paulrigor we can also change the installation location if that's worth trying

PR #112, opened 2023-12-28
Do not ouptut individual load parameters if not in original network

### LINKED ISSUE #109: Issues with individual load shedding costs
**Issue type**
- [ x] Bug

**Relates to**
- [ x] OPFLOW

**Summary**

OPFLOW load shedding costs can be specfied as a blanket cost for any load reduction (using `-opflow_loadloss_penality`) or per-load costs read from the ExaGO-specific `loadcost` array in network input.  

OPFLOW MATPOWER output always contains individual load costs even if they were not specified that way.  This should not be the case.  The `loadcost` field should not appear in the output if it did not appear in the input.  

Also, some problems have been observed with the initialization of the costs individual loads.  If the `loadcost` field does not appear in the network input, individual load costs should be initialized to the blanket cost.

wperkins:
**Merge request type**
- [x] Resolves bug

**Relates to**
- [x] OPFLOW

**This MR updates**
- [x] Header files
- [x] Source code

**Summary**

Load shedding parameters can be individually by each load using the `mpc.loadcost` field in the MATPOWER network.  These individual load loss parameters were added to the output even if they were never specified or used.  This fixes that. Individual load shedding parameters are output only if they were read from the input network. Closes #109.  

If load shedding parameters are not read from the input network, they are initialized with the OPFLOW blanket cost (`-opflow_loadloss_penality`), but only if needed (`-opflow_include_loadloss_variables`).

### DIFF
--- a/include/private/psimpl.h
+++ b/include/private/psimpl.h
@@ -451,6 +451,8 @@ struct _p_PS {
 
   PSSystemSummary sys_info;
 
+  PetscBool read_load_cost;     /* are individual load costs assigned? */
+
   PetscBool setupcalled; /* Is setup called on PS? */
 
   PetscLogDouble solve_real_time;
--- a/include/private/psimpl.h
+++ b/include/private/psimpl.h
@@ -451,7 +451,7 @@ struct _p_PS {
 
   PSSystemSummary sys_info;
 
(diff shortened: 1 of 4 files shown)


abhyshr (approved):
This is good to merge.

PR #131, opened 2024-03-26
TCOPFLOW Python Bindings First Pass

cameronrutherford:
**Merge request type**
- [X] New feature
- [ ] Resolves bug
- [ ] Documentation
- [ ] Other

**Relates to**
- [ ] OPFLOW
- [ ] SOPFLOW
- [ ] SCOPFLOW
- [X] TCOPFLOW
- [ ] CMake build system
- [ ] Spack configuration
- [ ] Manual
- [ ] Web docs
- [ ] Other

**This MR updates**
- [ ] Header files
- [ ] Source code
- [ ] CMake build system
- [ ] Spack configuration
- [ ] Web docs
- [ ] Manual
- [X] Other - python bindings

**Summary**

Summarize the MR concisely

**Linked Issue(s)**

Link any relevant issues. If merging into the default branch, any directly addressed issues will be closed when merging into the default branch (develop).

Since GitHub is specific on language, make sure to include direct language if you want to have the issue closed:
- Closes #issue
- Resolves #issue

You can also reference issues, but this will not close the issue when the merge request is merged:
- Partly related to #issue
- #issue
- Partially addresses #issue


Jayapreethi:
Hi @cameronrutherford, this is ready for review. Please let me know about any further required changes .

interfaces/python/CMakeLists.txt
```diff
@@ -18,8 +18,10 @@ if(EXAGO_ENABLE_IPOPT OR EXAGO_ENABLE_HIOP)
 endif()
 
 if(EXAGO_ENABLE_IPOPT)
-  list(APPEND EXAGO_PY_SRCS exago_python_scopflow.cpp exago_python_sopflow.cpp)
-  list(APPEND EXAGO_PY_LIBS ExaGO::SOPFLOW ExaGO::SCOPFLOW)
+  list(APPEND EXAGO_PY_SRCS exago_python_scopflow.cpp exago_python_sopflow.cpp
+       exago_python_tcopflow.cpp
```
cameronrutherford, line 22: Does tcopflow depend on Ipopt being available? @abhyshr

interfaces/python/example_tcopflow.py
cameronrutherford, line 1: This is a blank file. Is it meant to have contents similar to `interfaces/python/example_<app>.py`?

tests/interfaces/python/test_5_tcopflow.py
```diff
@@ -0,0 +1,40 @@
...
+from mpi4py import MPI
+import tempfile
+import os
+import shutil
+import pytest
+from check_preconditions import check_preconditions
+import mpi4py.rc
+mpi4py.rc.threads = False
+
+check_preconditions()
```
cameronrutherford:
```suggestion
from mpi4py import MPI
import tempfile
import os
import shutil
import pytest
from check_preconditions import check_preconditions
import mpi4py.rc
mpi4py.rc.threads = False

check_preconditions()
import exago
```

@Jayapreethi can you try out this change?

You had an error where you were running something along the lines of:

- `cmake -B build -S .`
- `cmake --build build -- -j 8`
- `cd build && ctest -VV -R Python`

And were getting errors with `import exago`. The whole point of the `check_preconditions()` functions seems to be to actually print some helpful debug messages, so I think moving that to before `import exago` would make the error message much more friendly.

Please also extend this change into other tests if you verify that it helps and works as I have described.

tests/interfaces/python/test_5_tcopflow.py
```diff
@@ -0,0 +1,40 @@
...
+
+exago_ignore = -1000000
```
cameronrutherford: Unused?
```suggestion

```

tests/interfaces/python/test_5_tcopflow.py
```diff
@@ -0,0 +1,40 @@
...
+    assert isinstance(obj, float)
+
+    n = tcopf.get_num_iterations()
+    assert isinstance(n, int)
+
+
+@pytest.mark.nocomm
+@pytest.mark.MPI
+def test_tcopflow_1():
+    run_tcopflow('IPOPT')
```
cameronrutherford, line 37:
```suggestion
@pytest.mark.nocomm
@pytest.mark.MPI
def test_tcopflow_1():
    tcopf = exago.TCOPFLOW()
    path = exago.prefix()

    tcopf.set_tolerance(1.0E-03)
    tcopf.set_network_data(os.path.join(
        path, 'share', 'exago', 'datafiles', 'case9', 'case9mod_gen3_wind.m'))
    tcopf.set_solver('IPOPT')

    tcopf.setup()
    tcopf.solve()

    assert type(tcopf.get_convergence_status()) == bool

    obj = tcopf.get_objective()
    assert isinstance(obj, float)

    n = tcopf.get_num_iterations()
    assert isinstance(n, int)
```

I don't see the need to abstract our the `run_tcopflow` function here. In #140, make sure to name each test to be unique to the function being tested. Arguably, this function should be split up to test the usage of `get_convergence_status`, `get_objective` and `get_num_iterations` separately.

I know that our test cases are always going to converge, but I think asserting the function returns a bool is more sensible, and perhaps we also want a dummy problem that does not converge to validate we can return that correctly as well...

cameronrutherford (commented):
Looking good! Requested some changes but this is almost ready to merge

tests/interfaces/python/test_5_tcopflow.py
jaelynlitz: It would be preferable to test each function individually. However, like with [sopflow](https://github.com/pnnl/ExaGO/blob/develop/tests/interfaces/python/test_4_sopflow.py) there are issues with the constructor/destructor. Simply creating a `tcopf` object then destroying it causes a seg fault bc some memory isn't allocated until `setup` is run. I suppose we could look into writing unit tests instead with calls to `setup` before testing each function but I opted to have us follow the sopflow framework first.

cameronrutherford: Okay in digging through source code, it seems like we might need to re-visit `sopflow` python testing since we should be able to run `sopf = exago.SOPFLOW(); sopf.setup()` without causing too many issues... We should either fix this in an issue, or document it slightly better and explain...

For `tcopflow` though, I think writing tests such that `setup` is called isn't too bad. Perhaps there should also be an issue created though for fixing our destructors to not crash if people create and release an object without calling `setup`? Regardless we can figure that out in #140

cameronrutherford: For more complex applications I think it makes sense to have an abstracted `run_tcopflow` to reduce copy and paste when we want to run a few times against a set of different configurations, but I think that can live alongside more "unit" tests while keeping the "functionality" tests.

interfaces/python/example_tcopflow.py
```diff
@@ -0,0 +1,17 @@
...
+exago.initialize("tcopflow_test")
+tcopf = exago.TCOPFLOW()
+path = exago.prefix()
+tcopf.set_network_data(
+    os.path.join(path, 'share', 'exago', 'datafiles', 'case9', 'case9mod.m'))
+tcopf.set_solver("IPOPT")
+tcopf.solve()
+tcopf.print_solution(0)
+
+# Delete instance before finalization (try, at least)
```
cameronrutherford, line 14: wdym try, at least? Does this not work when running?

tests/interfaces/python/test_0_initialize.py
```diff
@@ -1,10 +1,12 @@
 # test initializing exago only once
+import exago
```
cameronrutherford:
```suggestion
```
Don't `import exago` before `check_preconditions`

tests/interfaces/python/test_5_tcopflow.py
```diff
@@ -0,0 +1,38 @@
+import exago
```
cameronrutherford: Move this import to after `check_preconditions()` is called.

Also, make sure to extend this fix to all of our other tests

cameronrutherford (commented):
Please make sure as well to update the python interface markdown documentation @Jayapreethi

tests/interfaces/python/test_5_tcopflow.py
cameronrutherford: #142 and #140 should cover this

interfaces/python/CMakeLists.txt
cameronrutherford: @abhyshr assuming that it does need IPOPT. Feel free to clarify otherwise...

cameronrutherford: Maybe we need an issue to update TCOPFLOW docs with this information as well?

tests/interfaces/python/test_5_tcopflow.py
cameronrutherford: We navigated this offline... Still need to apply to all files

jaelynlitz (changes requested):
LGTM - please add the functions that you've ported in a new table for TCOPFLOW in [this doc page](https://github.com/pnnl/ExaGO/blob/develop/docs/python_bindings.md#sopflow) then I think this is good to go

Jayapreethi:
Documented the tested functions of TCOPFLOW.

tests/interfaces/python/test_0_initialize.py
```diff
@@ -1,10 +1,11 @@
 # test initializing exago only once
+from check_preconditions import check_preconditions
 import os
 import pytest
-from check_preconditions import check_preconditions
 import mpi4py.rc
 mpi4py.rc.threads = False
 from mpi4py import MPI  # noqa
+check_preconditions()
 import exago  # noqa
```
cameronrutherford, line 11: Eliminate duplicate `check_preconditions()`

tests/interfaces/python/test_1_pflow.py
```diff
@@ -1,11 +1,13 @@
...
 import pytest
-from check_preconditions import check_preconditions
 import mpi4py.rc
 mpi4py.rc.threads = False
 from mpi4py import MPI  # noqa
+check_preconditions()
 import exago  # noqa
 
+
 check_preconditions()
```
cameronrutherford:
```suggestion

```

tests/interfaces/python/test_2_opflow.py
```diff
@@ -1,9 +1,9 @@
...
 import os
 import pytest
-from check_preconditions import check_preconditions
 import mpi4py.rc
 mpi4py.rc.threads = False
-from mpi4py import MPI  # noqa
+check_preconditions()
 import exago  # noqa
 
 check_preconditions()
```
cameronrutherford:
```suggestion
```

tests/interfaces/python/test_3_scopflow.py
```diff
@@ -1,11 +1,12 @@
...
 import shutil
 import pytest
-from check_preconditions import check_preconditions
 import mpi4py.rc
 mpi4py.rc.threads = False
 from mpi4py import MPI  # noqa
+check_preconditions()
 import exago  # noqa
 
 check_preconditions()
```
cameronrutherford:
```suggestion
```

tests/interfaces/python/test_4_sopflow.py
```diff
@@ -1,13 +1,15 @@
...
 import pytest
-from check_preconditions import check_preconditions
 import mpi4py.rc
 mpi4py.rc.threads = False
 from mpi4py import MPI  # noqa
+check_preconditions()
 import exago  # noqa
 
+
 check_preconditions()
```
cameronrutherford:
```suggestion
```

tests/interfaces/python/test_5_tcopflow.py
```diff
@@ -0,0 +1,39 @@
...
+from mpi4py import MPI
+import tempfile
+import os
+import shutil
+import pytest
+import mpi4py.rc
+mpi4py.rc.threads = False
+check_preconditions()
+
+check_preconditions()
```
cameronrutherford:
```suggestion
```

tests/interfaces/python/test_9_finalize.py
```diff
@@ -1,7 +1,9 @@
 # test only run once things here
-import pytest
 import exago
+import pytest
 from check_preconditions import check_preconditions
+check_preconditions()
+
 
 check_preconditions()
```
cameronrutherford:
```suggestion
```

tests/interfaces/python/test_5_tcopflow.py
cameronrutherford: Please make sure to move this import.........

cameronrutherford (commented):
Minor cleanup

tests/interfaces/python/test_9_finalize.py
```diff
@@ -1,8 +1,7 @@
 # test only run once things here
-import pytest
 import exago
```
cameronrutherford:
```suggestion

```

tests/interfaces/python/test_9_finalize.py
```diff
@@ -1,8 +1,7 @@
...
+import pytest
 from check_preconditions import check_preconditions
-
 check_preconditions()
```
cameronrutherford, line 4:
```suggestion
check_preconditions()
import exago
```

cameronrutherford (commented):
One final missed `check_preconditions()`, apart from that looks good to merge!

cameronrutherford:
Spack ubuntu pipelines are failing... See #144

@jaelynlitz or @abhyshr with one final review I think this is good to merge! Thanks @Jayapreethi

abhyshr (approved):

PR #141, opened 2024-05-03
Update deceptionVariables.sh

cameronrutherford:
**Merge request type**
- [x] Resolves bug

**Relates to**
- [x] CMake build system

**This MR updates**
- [x] Other

**Summary**

When people run our build script from the wrong directory, `SRCDIR` can be configured incorrectly. This makes sure the variable isn't exported and should fix the bug...


cameronrutherford:
Surprisingly Incline was one of the few platforms that didn't have this bug...

jaelynlitz (approved):

PR #144, opened 2024-05-17
Bugfix spack ubuntu CPU build

cameronrutherford:
Adds `--fail-fast` to the spack install to avoid the installation running for any longer than it needs to

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -13,6 +13,17 @@ env:
 on: [pull_request]
 
 jobs:
+  cleanup_registry:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Cleanup
+        uses: actions/delete-package-versions@v5
+        with: 
+          package-name: 'exago'
+          package-type: 'container'
+          min-versions-to-keep: 100
+          delete-only-untagged-versions: 'true'
+          
   base_image_build:
     runs-on: ubuntu-22.04
     permissions:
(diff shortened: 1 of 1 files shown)


cameronrutherford:
Noticed the ubuntu build was failing in #131, so this is an attempt to speed up the rate of failure in those builds....

cameronrutherford:
Not really sure why this build is now failing... Ideas:

- [ ] Use a spack provided base image
- [ ] Pin spack version
- [ ] Re-use packages more aggressively

Still really unclear why this is happening since it's an error in getting the C compiler to even do anything

.github/workflows/spack_cpu_build.yaml
```diff
@@ -116,31 +120,38 @@ jobs:
...
+              build_stage: $SPACK_CACHE/build_stage
               install_tree:
                 root: /opt/spack
                 padded_length: False
             mirrors:
               local-buildcache: oci://${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
               spack: https://binaries.spack.io/develop
+            packages:
+              all:
+                require: "%gcc"
```
cameronrutherford, line 135: For some reason, clang was getting picked up by default as the compiler, and this caused spack to fail horribly. This seems to be all that was necessary to fix the issue

cameronrutherford:
Ready for review + merge after pipelines pass. Since this uses a newer gcc and ubuntu version, the whole build is happening from scratch this first time. Future runs should be quicker

cameronrutherford:
All green check marks, so I will merge once review is finished.

jaelynlitz (approved):

PR #149, opened 2024-06-28
Update Summit build scripts with new module dependencies.

pelesh:
**Merge request type**
- Resolves bug

**Relates to**
- CMake build system
- Spack configuration

**This MR updates**
- CMake build system

**Summary**

Script loading modules with ExaGO dependencies on Summit has been out of date. This PR updates build scripts to point to current modules available on Summit built by @nkoukpaizan.

**Linked Issue(s)**


nkoukpaizan:
This will need a bit of work for the develop branch. Some of the [module I built](https://github.com/pnnl/ExaGO/tree/nicholson/parco-summit-alpine2) are older versions of the dependencies (e.g., HiOp v0.7.2; the interface has changed since).

pelesh:
> This will need a bit of work for the develop branch. Some of the [module I built](https://github.com/pnnl/ExaGO/tree/nicholson/parco-summit-alpine2) are older versions of the dependencies (e.g., HiOp v0.7.2; the interface has changed since).

Is there any major dependency other than HiOp where the new version changes the API? If not, I would just remove HiOp module from the script, since in most use cases one would want to build their own HiOp. There is no CI pipeline using these modules so the testing would work as well as it works now.

nkoukpaizan:
Yeah, the script currently doesn't load the HiOp module, which is fine when providing our own HiOp build. But this breaks the default ExaGo buildsystem (i.e. running `MY_CLUSTER=summit ./buildsystem/build.sh`). Merging it as is might come back to bite us later.

nkoukpaizan (approved):
We've had confirmation from multiple sources that this currently works on Summit. I'll create another PR for ENG142. 

We'll need @cameronrutherford's review to merge, as I was the last one to push.

buildsystem/spack/summit/spack.yaml
```diff
@@ -1,11 +1,12 @@
 spack:
   specs:
   - exago@develop%gcc@10.2.0+cuda+mpi+hiop+raja+ipopt+python cuda_arch=70 build_type=Release
-    ^hiop@0.7.2+cuda+mpi+raja+kron+sparse+ginkgo+cusolver_lu cuda_arch=70
+    ^hiop@develop+cuda+mpi+raja+kron+sparse+ginkgo+cusolver_lu cuda_arch=70
+    ^cub@1.16.0
```
cameronrutherford, line 5: Is this cub dependency summit specific, or should this be expressed in our package.py somehow?

cameronrutherford (approved):
Just one small comment, but otherwise we can merge this quickly

buildsystem/spack/summit/spack.yaml
nkoukpaizan: Good catch. It's no longer needed on Summit either. There was a compatibility issue between cub and the old cuda modules on Summit, so I had experimented with different cub versions. Now, I'm letting Spack build cuda, so I no longer need to enforce a specific cub version. If you don't mind, I'll remove in the next PR where I'll install the modules in ENG142 directories.

cameronrutherford: Great

PR #89, opened 2023-11-27
Build with rocm/5.6 on Frontier (~ginkgo)

nkoukpaizan:
**Merge request type**
- [x] New feature
- [ ] Resolves bug
- [ ] Documentation
- [ ] Other

**Relates to**
- [ ] OPFLOW
- [ ] SOPFLOW
- [ ] SCOPFLOW
- [ ] TCOPFLOW
- [ ] CMake build system
- [x] Spack configuration
- [ ] Manual
- [ ] Web docs
- [ ] Other

**This MR updates**
- [ ] Header files
- [ ] Source code
- [ ] CMake build system
- [x] Spack configuration
- [ ] Web docs
- [ ] Manual
- [ ] Other

**Summary**

This MR updates the Spack configuration and the corresponding modules on Frontier to build with rocm/5.6.
A few notes:
- Building for `x86-64` target instead of `zen3` because of a compiler bug raised by PETSc (see #65). I opened an OLCF ticket to address this.
- Despite correctly building for x86-64, the variable $(spack arch) still returns `zen3`, so the path to the modules files was not correctly updated by default. I manually modified it, and I welcome suggestions for a sustainable solution.
- rocm/5.6 now enforces the more recent paths for header files (see https://rocm.docs.amd.com/en/docs-5.6.0/CHANGELOG.html#id31). This came up when attempting to build ExaGO in `Debug` mode against magma@2.6.2, which was using the old paths. magma@2.7.1+ implements a logic to ensure the correct paths are used depending on the ROCm version (see https://bitbucket.org/icl/magma/commits/f07ddb12ca3f3aad9253342b6800e48bd8f02888). The current build on Frontier relies on magma@2.7.2.

Issue #127, opened 2024-03-13
ExaGO fails to build: Missing HSL


pelesh:
When building with clang I get following link error:
```
[ 56%] Linking CXX executable opflow
ld.lld: error: undefined symbol: mc19ad_
>>> referenced by IpEquilibrationScaling.cpp
>>>               IpEquilibrationScaling.o:(Ipopt::EquilibrationScaling::DetermineScalingParametersImpl(Ipopt::SmartPtr<Ipopt::VectorSpace const>, Ipopt::SmartPtr<Ipopt::VectorSpace const>, Ipopt::SmartPtr<Ipopt::VectorSpace const>, Ipopt::SmartPtr<Ipopt::MatrixSpace const>, Ipopt::SmartPtr<Ipopt::MatrixSpace const>, Ipopt::SmartPtr<Ipopt::SymMatrixSpace const>, Ipopt::Matrix const&, Ipopt::Vector const&, Ipopt::Matrix const&, Ipopt::Vector const&, double&, Ipopt::SmartPtr<Ipopt::Vector>&, Ipopt::SmartPtr<Ipopt::Vector>&, Ipopt::SmartPtr<Ipopt::Vector>&)) in archive /lustre/orion/eng145/world-shared/spack-install/linux-sles15-x86_64/clang-16.0.0-rocm5.6.0-mixed/ipopt-3.12.10-7fp33q627rou44fzquk57llhwoqqeuho/lib/libipopt.a
```
It seems HSL is not found. This is for clean checkout of `develop` on Frontier. I can seeHSL module loaded:
```
$ ml

Currently Loaded Modules:
...
 34) coinhsl/2019.05.21-gcc-12.2.0-mixed-and6kty
 35) hipblas/5.6.0-clang-16.0.0-rocm5.6.0-mixed-pgkobjo
```
I used following command to build (all defaults):
```
$ CC=clang CXX=clang++ FC=flang cmake ../exago
$ make
```

_Originally posted by @pelesh in https://github.com/pnnl/ExaGO/issues/126#issuecomment-1995805427_

pelesh:
Additional info -- if I enable GPU support, this problem is gone. It looks as if there is a bug in CMake that enables Fortran support only if GPU support is enabled, as well.

CC @cameronrutherford @nkoukpaizan

pelesh:
Can't reproduce this anymore. Closing.

PR #151, opened 2024-07-29
Nicholson/frontier worldshared cpe23.12

nkoukpaizan:
**Merge request type**
- [ ] New feature
- [x] Resolves bug
- [ ] Documentation
- [ ] Other

**Relates to**
- [ ] OPFLOW
- [ ] SOPFLOW
- [ ] SCOPFLOW
- [ ] TCOPFLOW
- [ ] CMake build system
- [x] Spack configuration
- [ ] Manual
- [ ] Web docs
- [ ] Other

**This MR updates**
- [ ] Header files
- [ ] Source code
- [ ] CMake build system
- [x] Spack configuration
- [ ] Web docs
- [ ] Manual
- [ ] Other

**Summary**
Previously built ExaGO dependencies on Frontier no longer work after an update to the system software environment. This PR uses a more recent programming environment with the appropriate module loaded. 

I documented test failures (similar to previous observations in #89) in issue #152.

### DIFF
--- a/buildsystem/clang-hip/crusher/base.sh
+++ b/buildsystem/clang-hip/crusher/base.sh
@@ -6,12 +6,13 @@ export PROJ_DIR=/autofs/nccs-svm1_proj/eng145
 module reset
 
 # System modules
-module load PrgEnv-amd
+module load PrgEnv-gnu-amd
 module load cpe/23.12
 module load craype-x86-trento
 module load craype-accel-amd-gfx90a
-module load amd/5.7.1
+module load amd-mixed/5.7.1
 module load rocm/5.7.1
+module load gcc-native/12.3
 module load cray-mpich/8.1.28
 module load libfabric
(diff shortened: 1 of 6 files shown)


cameronrutherford:
LGTM but we should figure out the path forward for #152 before merging this IMO

pelesh (approved):
I was able to build and run ExaGO with new dependency management configuration. 

I suggest we merge. Issues #152 and #127 are out of scope of this PR and should be addressed separately.

pelesh:
> LGTM but we should figure out the path forward for #152 before merging this IMO

#152 does not seem to be caused by this PR, it is just exposed now that we can build and run tests on Frontier. Merging this could speed up debugging and will not affect tests on existing CI pipelines.

cameronrutherford:
@pelesh I agree, so I merged :)

PR #156, opened 2024-09-03
Add check for existing base image to speedup pipelines.

### LINKED ISSUE #153: Don't build custom base image for ExaGO in Spack pipelines, and add `spack test`
**Issue type**
- [x] New feature

**Relates to**
- [x] Other

**Summary**

https://github.com/ORNL/ReSolve/pull/181 - this PR implements the suggestion. It will moderately speedup all spack pipelines ran in all PRs.

https://github.com/ORNL/ReSolve/blob/develop/.github/workflows/spack_cpu_build.yaml#L138 - testing support also shouldn't bee too difficult to add in addition to this.

### LINKED ISSUE #155: Add `spack test` support to the GitHub spack pipelines
https://github.com/ORNL/ReSolve/pull/118/files - would be a simple change similar to what is done in this Re::Solve PR, and would ensure that each build of ExaGO we do runs and passes tests in the GitHub CI

cameronrutherford:
Closes #153 and closes #155

### DIFF
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -152,7 +152,7 @@ jobs:
                 padded_length: False
             mirrors:
               local-buildcache: oci://${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-              spack: https://binaries.spack.io/develop
+            # spack: https://binaries.spack.io/develop
             packages:
               all:
                 require: "%gcc"
--- a/.github/workflows/spack_cpu_build.yaml
+++ b/.github/workflows/spack_cpu_build.yaml
@@ -156,6 +156,7 @@ jobs:
             packages:
               all:
                 require: "%gcc"
+          EOF
 
       - name: Configure GHCR mirror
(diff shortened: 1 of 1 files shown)


cameronrutherford:
@nychiang I proceeded to accidentally reproduce and then proceed to fix the same issue that we were seeing in the HiOp spack pipelines. This is promising for fixing stuff there (I hope).

.github/workflows/spack_cpu_build.yaml
```diff
@@ -129,7 +152,7 @@ jobs:
                 padded_length: False
             mirrors:
               local-buildcache: oci://${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-              spack: https://binaries.spack.io/develop
+            # spack: https://binaries.spack.io/develop
```
cameronrutherford, line 154: Have to avoid using the public spack binary mirror for now due to https://github.com/spack/spack/issues/43919

cameronrutherford:
Looks like no tests are being run... Might have to make some modifications to the ExaGO package to fix this, but we can debug that in a different PR...

Issue #188, opened 2025-12-17
Finish PSSE parser


PhilipFackler:
- [x] Add switched shunts (with place-holders for remaining record types)
- [x] Ensure compliance with v34
- [x] Emit warning in conversion function when elements are present that are not supported by `PS`
- [x] Handle optional inputs
- [ ] Add remaining record types
  - [ ] system-wide data
  - [x] area
  - [ ] two-terminal dc
  - [ ] vsc dc
  - [ ] impedence correction
  - [ ] multi-terminal dc
  - [ ] multi-section line
  - [x] zone
  - [ ] inter-area transfer
  - [x] owner
  - [ ] facts device
  - [ ] gne
  - [ ] induction machine

pelesh:
CC @hambrickjc @tsybinae

nkoukpaizan:
This has been created on the new [repo](https://github.com/ORNL/ExaGO/issues/2). Closing here.

PR #194, opened 2026-01-15
Iterations on PSSE parser

PhilipFackler:
**Merge request type**
- [X] New feature
- [X] Resolves bug
- [ ] Documentation
- [ ] Other

**Relates to**
- [ ] OPFLOW
- [ ] SOPFLOW
- [ ] SCOPFLOW
- [ ] TCOPFLOW
- [ ] CMake build system
- [ ] Spack configuration
- [ ] Manual
- [ ] Web docs
- [X] Other

**This MR updates**
- [X] Header files
- [X] Source code
- [ ] CMake build system
- [ ] Spack configuration
- [ ] Web docs
- [ ] Manual
- [ ] Other

**Summary**

Multiple improvements to PSSE parser:
- Initial v34 support with caveats
  - "system-wide data" block is skipped for now
  - Still not parsing every record type
- Use more consistent messages for errors and logging (including warning for elements not supported by `PS`)
- Handle record items that are optional in the specification
- Resolve defaults that depend on data in other record types (e.g., `load.area` defaults to area of bus referenced by `load.i`)

**Linked Issue(s)**
- #188


tsybinae:
Applied ./pflow -netfile for three different files, including Hawaii 40 https://electricgrids.engr.tamu.edu/hawaii40/.
Produces this error message.
```
 ~/lib/ExaGO/build/applications$ ./pflow -netfile Hawaii40_20231026.RAW
terminate called after throwing an instance of 'std::runtime_error'
  what():  Cannot resolve bus id
[LAP136650:03927] *** Process received signal ***
[LAP136650:03927] Signal: Aborted (6)
[LAP136650:03927] Signal code:  (-6)
[LAP136650:03927] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x45330)[0x7333d6845330]
[LAP136650:03927] [ 1] /lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0x11c)[0x7333d689eb2c]
[LAP136650:03927] [ 2] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x1e)[0x7333d684527e]
[LAP136650:03927] [ 3] /lib/x86_64-linux-gnu/libc.so.6(abort+0xdf)[0x7333d68288ff]
[LAP136650:03927] [ 4] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xa5ff5)[0x7333d6ca5ff5]
[LAP136650:03927] [ 5] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xbb0da)[0x7333d6cbb0da]
[LAP136650:03927] [ 6] /lib/x86_64-linux-gnu/libstdc++.so.6(_ZSt10unexpectedv+0x0)[0x7333d6ca5a55]
[LAP136650:03927] [ 7] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xbb391)[0x7333d6cbb391]
[LAP136650:03927] [ 8] ./pflow(+0x108f2)[0x5b632f2b88f2]
[LAP136650:03927] [ 9] ./pflow(+0x2a342)[0x5b632f2d2342]
[LAP136650:03927] [10] ./pflow(+0x2aa86)[0x5b632f2d2a86]
[LAP136650:03927] [11] ./pflow(+0x2ad92)[0x5b632f2d2d92]
[LAP136650:03927] [12] ./pflow(+0x2b539)[0x5b632f2d3539]
[LAP136650:03927] [13] ./pflow(+0x1693e)[0x5b632f2be93e]
[LAP136650:03927] [14] ./pflow(+0x1248f)[0x5b632f2ba48f]
[LAP136650:03927] [15] /lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x7333d682a1ca]
[LAP136650:03927] [16] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x7333d682a28b]
[LAP136650:03927] [17] ./pflow(+0x15065)[0x5b632f2bd065]
[LAP136650:03927] *** End of error message ***
Aborted (core dumped)
```

pelesh:
Here is what I am getting with the updated PSSE parser.

The case `Hawaii40_20231026.RAW` should be parsed correctly but it returns an error:
```
$ ./pflow -netfile Hawaii40_20231026.RAW 
[ExaGO] Parsing PSS(R)E raw data file: 'Hawaii40_20231026.RAW'
[ExaGO] Generator at bus : voltage setpoint (1) different from bus voltage magnitude (0.991225)
terminate called after throwing an instance of 'ExaGOError'
  what():  Generator at bus : voltage setpoint (1) different from bus voltage magnitude (0.991225)
Aborted (core dumped)
```

The file that should trigger failure, `bench.raw`, does that and the parser correctly recognizes unsupported format:
```
$ ./pflow -netfile bench.raw 
[ExaGO] Parsing PSS(R)E raw data file: 'bench.raw'
[ExaGO] PSS(R)E Power Flow Raw Data Files are supported only for versions 32, 33 and 34. This file is tagged version 31
terminate called after throwing an instance of 'ExaGOError'
  what():  PSS(R)E Power Flow Raw Data Files are supported only for versions 32, 33 and 34. This file is tagged version 31
Aborted (core dumped)
```

**Question**: Why are we throwing an exception after reading poorly structured input? We could just return.

src/ps/psse.cpp
```diff
@@ -71,230 +113,360 @@ class IntOrStringParse {
...
+        Error("PSS(R)E parser: Encountered duplicate bus number: " +
+              std::to_string(buses_[i].i));
       }
     }
     {
       auto [_, ins] = name_map_.emplace(buses_[i].name, i);
       if (!ins) {
-        throw std::runtime_error("Bus names non-unique");
+        Error("PSS(R)E parser: Encountered duplicate bus name: " +
+              buses_[i].name);
```
tsybinae, line 497: Bus numbers have to be unique, bus names can be not unique, because they carry the names of towns or cities, and if two different states have a town called “Gainesville”, then you can have bus 123456 “Gainesville” and 678901 “Gainesville”, suggest to remove the check for unique bus names.

src/ps/psse.cpp
```diff
@@ -544,9 +795,8 @@ PetscErrorCode ConvertToPS(PS ps, const Network &nw) {
     }
     auto bus_ii = nw.bus_mapping.GetInternalIndex(shunt.i);
     if (ps->bus[bus_ii].nshunt > 0) {
-      throw std::runtime_error(
-          "Bus " + std::to_string(shunt.i) +
-          ": more than one fixed shunt at bus not supported");
+      Error("Bus " + std::to_string(shunt.i.id) +
+            ": more than one fixed shunt at bus not supported");
```
tsybinae: If a bus has more than one fixed shunt, it is ok. They have sh.id, so “SH1”, “SH2”.

src/ps/psse.cpp
```diff
@@ -597,10 +847,12 @@ PetscErrorCode ConvertToPS(PS ps, const Network &nw) {
...
-      if (dgen.vs != bus.vm) {
-        throw std::runtime_error(
-            "Generator " + sgen.id +
-            ": set point voltage different from bus voltage magnitude");
+      if (!Approx(dgen.vs, bus.vm)) {
+        std::stringstream ss;
+        ss << "Generator at bus " << bus.bus_i << std::fixed
+           << ": voltage setpoint (" << dgen.vs
+           << ") different from bus voltage magnitude (" << bus.vm << ")";
+        Error(ss.str());
```
tsybinae: Generator voltage setpoint does not match bus voltage, this is normal.

tsybinae:
A few questions not specific to changes:
1. Why did you use bus name as an equal identifier to bus number? In principle, only bus number is used as reference. Bus name does not have to exist and somewhere in the code you also assume that it can copy from bus number. 
2. dgen.pgs = (dgen.pb + dgen.pt) / 2.0; Midpoint start? 
3. auto configure_line = [&nw](auto &line). Suggest to check if tap, r, x <>0, because then you have Z and tap2, which should be > 0.

src/ps/psse.cpp
PhilipFackler: Sounds good. It looks like other components can refer to buses by name ("extended bus id") rather than number. That's why I had that requirement. So should I assume that all bus references in the file will use the bus number?

src/ps/psse.cpp
PhilipFackler: OK I carried this check from the original parser. I'll move it to the convert function as a warning.

src/ps/psse.cpp
PhilipFackler: This check was in the original parser. Should I remove it altogether?

src/ps/psse.cpp
pelesh: I believe that "extended bus id" is used mainly for output. For computations, I would assume that bus number is used as a unique bus id.

CC @abhyshr

src/ps/psse.cpp
pelesh: Physically, it is okay to have more than one shunt, we need to verify that the model supports multiple shunts. I think producing warning when multiple shunts are connected to a bus is a good solution for now.

src/ps/psse.cpp
pelesh: Perhaps turn this into a warning instead of an error for now? We need to verify that the current model supports this.

src/ps/psse.cpp
PhilipFackler: From the docs:

> 1.2. Extended Bus Names
On its Bus Data record, each bus is assigned a bus number and a 12 character alphanumeric name. When the bus names input option of activity READ is enabled, data fields designating buses on load, fixed shunt, generator, non-transformer branch, transformer, area, two-terminal dc line, VSC dc line, multi-terminal dc line, multi-section line, FACTS device, switched shunt, GNE device, and induction machine, data records may be specified as either extended bus names enclosed in single quotes or as bus numbers.

What I hear you both saying is we should not allow this. All the listed component types should designate buses by integer id. Am I correct?

src/ps/psse.cpp
PhilipFackler: Changed to a warning

src/ps/psse.cpp
PhilipFackler: Changed to a warning

src/ps/psse.cpp
pelesh: I think this is correct.

@abhyshr, do you want to chime in?

PhilipFackler:
@tsybinae :

> 2. dgen.pgs = (dgen.pb + dgen.pt) / 2.0; Midpoint start?

I copied this logic from [here](https://github.com/pnnl/ExaGO/blob/5f7f528cb3e548a2db19bf77b7de29debdcb5be5/src/ps/psreaddata.cpp#L318-L322).

> 3. auto configure_line = [&nw](auto &line). Suggest to check if tap, r, x <>0, because then you have Z and tap2, which should be > 0.

This was also copied from the original reader. It combines the logic used for both branches and transformers [here](https://github.com/pnnl/ExaGO/blob/5f7f528cb3e548a2db19bf77b7de29debdcb5be5/src/ps/psreaddata.cpp#L337-L476). I see that `Zm` and `tap2` will always be >= 0 since they use squares. I'm not understanding what you're suggesting I check with `tapratio`, `R`, and `X`. Can you clarify?

My goal has been to reproduce the result of the original reader. Any changes needed will be for you experts to decide. :smiley_cat: And I'll be happy to implement them.

src/ps/psse.cpp
PhilipFackler: Per my discussion with @pelesh this is now contingent. Unique bus names are not required by default. But the requirement is triggered if any of the components listed in the above paragraph reference a bus with a string. I believe this PR is at a stable state to be merged.

hambrickjc:
Hawaii40 case successfully loads in pflow in both psse and matpower format. Application converges in same number of iterations but with slight differences between psse and matpower cases.

Hawaii40 case also successfully loads in opflow in both psse and matpower format, but does not converge in psse format which is expected due to lack cost information

Vmax and Vmin appear to be reversed when importing from raw files (see below...formatting is worse in a code block) but output correctly when importing from m files.

%% bus data
%	bus_i	type	Pd	Qd	Gs	Bs	area	Vm	Va	baseKV	zone	Vmax	Vmin	mult_Pmis	mult_Qmis	Pslack	Qslack
mpc.bus = [
	1	1	0	0	0	0	1	0.995064419	-1.20921355	138	1	0.9	1.1	0	0	0	0;
	2	2	60.7	0	0	0	1	1	-4.10106918	69	1	0.9	1.1	0	0	0	0;
	3	1	59.39	0	0	0	1	0.985795367	-4.73782564	69	1	0.9	1.1	0	0	0	0;
	4	1	22.47	0	0	0	1	0.982837787	-5.78191733	69	1	0.9	1.1	0	0	0	0;
	5	1	0	0	0	0	1	0.990705392	-2.14830164	138	1	0.9	1.1	0	0	0	0;
	6	1	27.46	0	0	0	1	0.984085723	-5.5620905	69	1	0.9	1.1	0	0	0	0;

PhilipFackler:
> Vmax and Vmin appear to be reversed when importing from raw files

@hambrickjc These were indeed swapped in my code. I fixed that just now.

I also removed the test comparing these values, because they were also swapped in the original parser:
Setting them [here](https://github.com/pnnl/ExaGO/blob/173ab38c5e55ac62f41196195dc51d3aa5724265/src/ps/psreaddata.cpp#L166-L167) then swapping them [here](https://github.com/pnnl/ExaGO/blob/173ab38c5e55ac62f41196195dc51d3aa5724265/src/ps/psreaddata.cpp#L231-L232).

hambrickjc (approved):
concerns have been addressed and the updated parser has been successfully tested against sample RAW files.

Issue #38, opened 2023-10-11
Transition dependencies from submodules into spack


cameronrutherford:
There are currently used when building ExaGO:
- [ ] toml11
- [ ] pybind11
- [ ] spdlog 

Now that one can completely disable logging, we should move to have other dependencies managed through spack, and then have them controlled by the relevant variants. We might have to control the version that we build with, but this would allow users to clone ExaGO without submodules and build accordingly.

This would have the additional side effect of remove the requirement to build ExaGO with submodules enabled in Spack, which should be reflected in [#391 (closed)](https://github.com/exasgd/frameworks/exago/-/issues/391)

pelesh:
@nkoukpaizan would you mind to chime in here as you are working on the updated spack package.

pelesh:
CC @PhilipFackler

nkoukpaizan:
I moved `spdlog` to an external dependency in #196 , since it is completely optional. We can remove the submodule in a subsequent PR. 

I have not worked on `pybind11` and `toml11` yet.

pelesh:
Addressed appropriately in #196

Issue #148, opened 2024-06-28
All but one ExaGO tests fail after clean build


pelesh:
**Issue type**
- Bug

**Relates to**
- CMake build system

**Summary**

After clean build all but One ExaGO tests fail. It seems there is a path issue somewhere. The error message is:
```
3% tests passed, 37 tests failed out of 38

Label Time Summary:
incline-skip    =   1.28 sec*proc (1 test)

Total Test time (real) =  48.13 sec

The following tests FAILED:
	  2 - UNIT_TESTS_OBJECTIVE_SIZE_OF_unittestx3.m_IPOPT_POWER_BALANCE_POLAR (Failed)
	  3 - UNIT_TESTS_OBJECTIVE_SIZE_OF_unittestx600.m_IPOPT_POWER_BALANCE_POLAR (Failed)
	  4 - UNIT_TESTS_OBJECTIVE_SIZE_OF_unittestx3.m_HIOPSPARSE_POWER_BALANCE_POLAR (Failed)
          ( ... )
	 38 - SOPFLOW_SOLUTION_OUTPUT_MINIMAL (Failed)
can't opendir /ccs/home/peles/src/exago/exago-github/build/tests/functionality: No such file or directory at /ccs/home/peles/src/exago/exago-github/tests/functionality/scripts/post_test.pl line 9.
Problem running command: perl /ccs/home/peles/src/exago/exago-github/tests/functionality/scripts/post_test.pl /ccs/home/peles/src/exago/exago-github/build/tests/functionality
Problem executing post-test command(s).
Errors while running CTest
``` 

**Exact commands to reproduce, if applicable**

```
git clone git@github.com:pnnl/ExaGO.git exago
mkdir build
cd build
CC=gcc CXX=g++ FC=gfortran cmake -DEXAGO_ENABLE_PYTHON=Off ../exago
make
ctest
```
I am using GCC 10.2, PETSc and other dependencies are built using ExaGO Spack package. Let me know what other debugging information you need me to provide.

pelesh:
I believe the issue here is that for some reason one must run `make install` before `make test`. Probably an issue in VMake configuration.

PR #196, opened 2026-01-29
Upgrade to Spack v1.1.1 and related fixes.

nkoukpaizan:
**Merge request type**
- [ ] New feature
- [ ] Resolves bug
- [ ] Documentation
- [x] Other

**Relates to**
- [ ] OPFLOW
- [ ] SOPFLOW
- [ ] SCOPFLOW
- [ ] TCOPFLOW
- [x] CMake build system
- [x] Spack configuration
- [ ] Manual
- [ ] Web docs
- [ ] Other

**This MR updates**
- [ ] Header files
- [ ] Source code
- [x] CMake build system
- [x] Spack configuration
- [ ] Web docs
- [ ] Manual
- [ ] Other

**Summary**

This upgrades the Spack submodule to v1.1.1 and deprecates incompatible build configurations (deception, newell and incline) that are not being tested with the new configuration.
- With the Spack upgrade, now pointing to upstream Spack and using local Spack repos for modified packages that have yet to be contributed upstream. I plan to submit appropriate PRs to spack-packages once we know this works.
- Updated Frontier modules accordingly. `romcc` is no longer a supported compiler, and mixes compiler modules are deprecated on Frontier. The only way to get a performant Fortran compiler for `coinhsl`, etc., and support HIP is to load `PrgEnv-gnu` and `rocm`. So, most packages are built built with`gcc` unless there is HIP code.
- Moved deprecated configurations to `deprecated` folders with corresponding `README` files.

**Linked Issue(s)**
- Partially addresses https://github.com/pnnl/ExaGO/issues/38 for spdlog


pelesh:
CC @cameronrutherford

nkoukpaizan:
@cameronrutherford Do you know a fix for the error I'm seeing?
```
Error: invalid push configuration for mirror local-buildcache: environment variable `***` (secret_variable) not set
```
Everything is building fine, but the push to the container registry is not working.

nkoukpaizan:
> @cameronrutherford Do you know a fix for the error I'm seeing?
> 
> ```
> Error: invalid push configuration for mirror local-buildcache: environment variable `***` (secret_variable) not set
> ```
> 
> Everything is building fine, but the push to the container registry is not working.

Finally figured out a way to make it work!

.github/workflows/spack_cpu_build.yaml
```diff
@@ -162,11 +169,14 @@ jobs:
       - name: Install
         run: spack -e . install --no-check-signature --fail-fast --keep-stage
 
-        #- name: Test Build
-        #  run: cd $(spack -e . location --build-dir exago@develop) && ctest -VV
+      #- name: Test Build
```
cameronrutherford: I can't remember why this is commented out.. perhaps there's an issue for it? Would be good to test things work before pushing.

buildsystem/clang-hip/frontier/base.sh
```diff
@@ -6,15 +6,20 @@ export PROJ_DIR=/autofs/nccs-svm1_proj/eng151
...
 module load cray-mpich
 module load libfabric
+module load cmake
+module load cray-python
 
-# Consider changing to $(which clang) as for deception
+# The CC, CXX, FC environment variables can be overriden by Spack's module generation.
+# The Frontier configuration was setup to avoid this override, but this is a
+# potential pitfall (e.g., when configuring another system). 
+# If all else fails, CMAKE_CXX_COMPILER superces these environment variables.
```
cameronrutherford:
```suggestion
# If all else fails, CMAKE_CXX_COMPILER suppresses these environment variables.
```
I'm not sure if this is the correction, but original has typo.

buildsystem/spack/load_spack.sh
```diff
@@ -57,6 +57,9 @@ source $base/env.sh &&
 	spack env create -d $SPACKENV $base/spack.yaml &&
 	spack env activate -p $SPACKENV &&
 
+        # Use local Spack repo for modified packages
+        spack repo add ${base}/../spack_repo/exago &&
```
cameronrutherford:
```suggestion
      # Use local Spack repo for modified packages
      spack repo add ${base}/../spack_repo/exago &&
```

Maybe I am just on mobile, but indentation looks off here.

buildsystem/spack/spack_repo/exago/packages/exago/__pycache__/package.cpython-311.pyc
cameronrutherford, line 1: Maybe we should gitignore and remove this 😄

buildsystem/spack/spack_repo/exago/packages/coinhsl/package.py
cameronrutherford, line 1: I suppose this is a nice convenience...

buildsystem/spack/spack_repo/exago/repo.yaml
```diff
@@ -0,0 +1,3 @@
+repo:
+  namespace: 'exago'
```
cameronrutherford, line 2: `spack install exago.exago`. Nice

buildsystem/spack/spack_repo/exago/packages/hiop/package.py
cameronrutherford, line 1: Also useful for testing changes not yet merged I presume

buildsystem/spack/spack_repo/exago/packages/hiop/__pycache__/package.cpython-311.pyc
cameronrutherford, line 1: Best to gitignore / remove this too

CMakeLists.txt
```diff
@@ -291,9 +291,8 @@ if(EXAGO_RUN_TESTS)
 endif()
 
 if(EXAGO_ENABLE_LOGGING)
-  add_subdirectory(${PROJECT_SOURCE_DIR}/tpl/spdlog/)
-  # -fPIC required for spdlog for use with python wrapper
-  set_property(TARGET spdlog PROPERTY POSITION_INDEPENDENT_CODE ON)
+  find_package(spdlog REQUIRED)
+  find_package(fmt REQUIRED)
```
cameronrutherford, line 305: Is the Python wrapper needing position independent code not relevant anymore?

Also, should we give a hint to the path here for the submodule?

Perhaps we only enforce that property here / in spack when Python is enabled.

buildsystem/spack/spack_repo/exago/packages/exago/package.py
```diff
@@ -0,0 +1,307 @@
...
+from spack_repo.builtin.build_systems.cuda import CudaPackage
+from spack_repo.builtin.build_systems.rocm import ROCmPackage
+
+from spack.package import *
+
+
+def submodules(package):
+    submodules = []
+    submodules.append("tpl/pybind11")
+    submodules.append("tpl/spdlog")
```
cameronrutherford, line 17: If you're making this an external dependency, then we should probably consider doing the following in pursuit of removing submodules (I think this is an existing issue / desirable goal):

1. Figure out a way to only clone some submodules for different versions of ExaGO. I imagine it's doable as `package` is passed to the function.
1. Add this as a dependency at the spack level (build time, link time?) for only certain versions.
1. Figure out how to express the dependency for PIC if it's necessary only for the Python wrapper(?)

This can also be another issue for later if you just want this merged and it works IMO.

buildsystem/spack/spack_repo/exago/packages/exago/package.py
cameronrutherford, line 1: Hard to discern the difference from our current `package.py`, so I just made the spdlog comment.

cameronrutherford (changes requested):
Some NITs, some other more interesting comments.

buildsystem/clang-hip/frontier/base.sh
nkoukpaizan: Addressed in [883a00b](https://github.com/pnnl/ExaGO/pull/196/commits/883a00b314bce58ee5e4a8162c6c1644bc5271ee).

buildsystem/spack/spack_repo/exago/packages/exago/__pycache__/package.cpython-311.pyc
nkoukpaizan: Good catch. It's already in the `.gitignore`, but I had to force add `spack_repo` and missed this. Addressed in [7d6f2fe](https://github.com/pnnl/ExaGO/pull/196/commits/7d6f2fe4f522d5da1573b6afa60b5a3143cc34aa).

buildsystem/spack/spack_repo/exago/packages/hiop/__pycache__/package.cpython-311.pyc
nkoukpaizan: Addressed in [7d6f2fe](https://github.com/pnnl/ExaGO/pull/196/commits/7d6f2fe4f522d5da1573b6afa60b5a3143cc34aa).

buildsystem/spack/spack_repo/exago/packages/exago/package.py
nkoukpaizan: I only made `spdlog` an external dependency, since that is truly optional. We can remove the submodule in a subsequent PR once the edits are propagated upstream.  We will also need to add patches to older versions when we make that change. The Spack-level dependency only applies to `develop` at the moment.

I have not addressed `pybind11` and `toml11`. One thing to consider it that making these external puts the burden on the user to have them installed.

CMakeLists.txt
nkoukpaizan: I'm requiring `spdlog` as an external dependency here, so I wouldn't use the submodule and `PIC` does not apply to the external target.

buildsystem/spack/spack_repo/exago/packages/hiop/package.py
nkoukpaizan: Not planning to keep it around long term. Only because I made edits that are not upstream.

buildsystem/spack/load_spack.sh
nkoukpaizan: Fixed in [5455ed1](https://github.com/pnnl/ExaGO/pull/196/commits/5455ed175b40752069091f29ff5af7f80ce87a43).

.github/workflows/spack_cpu_build.yaml
nkoukpaizan: The reason was that no tests were ever running in GH-Actions. We've been relying on manually testing on Frontier, and just checking that things build in GH-Actions. 

The Spack package relied on Spack's `--run-tests` to set `EXAGO_RUN_TESTS`. If I recall correctly, we can't used the built-in testing because of the need to install first. I've just added an independent `testing` option in the Spack package, independent of `--run-tests`, which revealed a number of failing tests. I'll have to look into it.

cameronrutherford:
LGTM if CI passes. I'm sure you'll get a better review from a maintainer about the `package.py` recipe anyway.

buildsystem/spack/spack_repo/exago/packages/exago/package.py
pelesh: I would suggest we discuss it a little more before we decide to treat toml and pybind as external packages. ExaGO already has too many dependencies and I agree with @nkoukpaizan that we need to minimize burden on the user.

buildsystem/spack/spack_repo/exago/packages/hiop/package.py
```diff
@@ -0,0 +1,334 @@
...
+        default=False,
+        description="Enable/Disable AXOM to use Sidre for scalable checkpointing",
+    )
+    variant("jsrun", default=False, description="Enable/Disable jsrun command for testing")
+    variant("shared", default=False, description="Enable/Disable shared libraries")
+    variant("mpi", default=True, description="Enable/Disable MPI")
+    variant("raja", default=False, when="@0.3.99:", description="Enable/Disable RAJA")
+    variant("kron", default=False, description="Enable/Disable Kron reduction")
+    variant("sparse", default=False, description="Enable/Disable Sparse linear algebra")
+    variant(
```
pelesh: I think HiOp should be built with sparse support by default. We are going to deprecate mixed dense-sparse module in ExaGO and PriDec and sparse HiOp modules will be pretty much all we'll need.

.github/workflows/spack_cpu_build.yaml
pelesh: I think we could fix easily the issue where we need to install before testing. As far as I remember, there was an issue about finding test input files and it was "fixed" by installing them first, and then accessing them in the installation directory. Also see #148.

buildsystem/spack/spack_repo/exago/packages/hiop/package.py
nkoukpaizan: That is the HiOp package, so I'll argue that the default value falls under the purview of HiOp. Not ExaGO.

pelesh: Actually, HiOp default is that sparse linear algebra is [on](https://github.com/llnl/hiop/blob/develop/CMakeLists.txt#L45). Perhaps the HiOp spack package has not been updated accordingly.

nkoukpaizan: Addressed in [721fc28](https://github.com/pnnl/ExaGO/pull/196/commits/721fc280666e0dfccff40729ae55c16a5a2173c9). I'll need to make a PR to upstream spack-packages and get it approved.

nkoukpaizan: On second thought, setting the default to True in the Spack package will likely fail CI in spack-packages, because it will depend on CoinHSL, which requires a licence, and I doubt that is setup. It doesn't hurt to try, though.

pelesh: We build Ipopt +mumps~mpi for GridKit, I believe and that works for testing. On the other hand, HiOp also has CoinHSL enabled by default, which indeed may be an issue for automated testing.

.github/workflows/spack_cpu_build.yaml
nkoukpaizan: These were just never tested (ran) in GH-Actions. I mean, there are things like trying to use 3 processors to run in parallel whereas the platform has maximum 2 cores.

nkoukpaizan: Just tested `make test` before `make installl` and it fails if I make sure to remove the existing `install` directory.

pelesh: Okay, I looked into failing tests when ExaGO is not installed, and I found the issue buried deep in test utilities. TOML files specify relative paths to input data files for tests. Setting [prefix](https://github.com/pnnl/ExaGO/blob/develop/tests/functionality/pflow/selfcheck.cpp#L114) for those paths is implemented in an error prone manner.  Actually, the path prefix for input files is hard-wired [here](https://github.com/pnnl/ExaGO/blob/develop/tests/functionality/toml_utils.h#L31).

CC @PhilipFackler

nkoukpaizan:
I fixed several issues related to running the tests in GH-Actions (e.g., detecting the maximum number of available cores for testing in parallel, fixing test dependencies in CMake when some EXAGO_ENABLE_* options are off). Now, we can also build against `ipopt~coinhsl+mumps ^mumps+mpi`. 

I documented the main remaining issues in #201. I suggest we keep the `Test build` stage of GH-Actions commented out for now.

pelesh (approved):

